PR #41, opened 2020-05-18
Create zh-hant.json

BlueWhaleYT:
Traditional Chinese Language which is translated by myself. I used a template zh-cn to start the work. I hope you will implement it.
Thanks.


deadlyjack:
Thank you so much brother for the translation. I really appreciate it.

PR #105, opened 2020-10-08
update zh-hant.json

Euxcbsks:
Change:
1. improve sentence smoothness
2. modify some translations
3. delete ["redo defination": "重做",] at line 72 in original file, because it's also deleted in en_us.json when I'm edit this file


BlueWhaleYT:
thx for inheriting my zh-hant translation and you are better than me to translate it. 😄

PR #110, opened 2020-10-21
Do not vibrate when vibration setting is turned off

diht2016:
Why disabling the "Vibrate on tap" setting does not remove vibrations in some cases (like long tapping on a word to select it)? Vibrations sometimes can be very annoying. So much annoying that they can force a person to fix this bug-like behavior and make a pull-request :)

(sorry for the newlines at the end of files, they were automatically added by github's built-in editor)


deadlyjack:
Thanks for fixing it.

PR #114, opened 2020-10-25
Update Japanese translation

paontv:


deadlyjack:
Thank you so much for updating the Japanese translation.

PR #117, opened 2020-10-31
cs-cz update and some corrections

nigdajznikec:
im sorry that i didnt react earlier, im so overwhelmed about all those python 'batteries'..
sometimes im forgeting even to go to sleep :D


nigdajznikec:
btw again: im the very noob on versioning source code, so please let me know if i'm doing something wrong or not as expected ;) thanks. Tom

nigdajznikec:
oh i see now :D .. ten commits to change one file.. sry i hoped that some 'magic in the middle' will consolidate those .. it looks like its time for man git or stop pushing every letter.. lol

deadlyjack:
Hi thank you so much for the update. Can you please fix the conflict.

PR #242, opened 2021-09-01
Add italian translation file

Stef-00012:


Stef-00012:
i'm remaking it since the file i translated is of an outdated version and are missing some texts

PR #342, opened 2022-02-08
Update ReactJS version to 17

BanDroid:
This version allow users to use hooks in React CDN.

How to use hooks:
```
import React from 'react';
function App() {
  React.useEffect(); // using useEffect hooks
  const [num, setNum] = React.useState(0) // using useState
  return (
    <div>
    </div>
  );
}
export default App;
```


deadlyjack (commented):
Why `<%name%> is loading...` is removed from index.html? Please add it and test it. After that, I'll merge it.

BanDroid:
> Why `<%name%> is loading...` is removed from index.html? Please add it and test it. After that, I'll merge it.

First of all, thank you for reviewing.

I just removed it because its unnecessary to have that loading paragraph shown up before showing the react project.

BanDroid:
Alright, i will update my commit 👌

deadlyjack:
Merged to main branch. Thank you for your contribution.

PR #344, opened 2022-02-11
Update helpers.js and react template.

BanDroid:
Now in react template, it support both **JS** and **JSX** file format.

**JSX** file format will automatically switch to mode-jsx.


deadlyjack:
Thanks for your contribution, I'll merge it after testing.

PR #346, opened 2022-02-14
Added Django, Javascript & c/cpp snippets and updated Beng language file.

deepkrg17:
## Django snippets
- for & if
- static or any tag
- block
- load

## JavaScript snippets
- addEventListener
- querySelector
- querySelectorAll

### Updated Bengali language translation file.


tanveer188:
actual it is not working

PR #402, opened 2022-05-23
Update constraint.js

DirtyApexAlpha:
Docs in npx rdme docs ./docs --version=v1.0 --key=HQq19auJDR4vTSBxNbpfvXQam6VooaTB
Changes handler api and signature name. Proper-Care(.pc) 
Constraint is never gonna happen for white jack-it throwable rdme-doc./docs@com


DirtyApexAlpha (commented):
Applications 'rdme' approve from feedbacks

Issue #410, opened 2022-06-04
F-Droid builds are out of date


balporig:
The current F-Droid version getting quite old `1.4.164 (164) - Added on 2021-10-12`

jstetten:
Very out-of-date. @deadlyjack, are you still planning to support F-droid in the future? Planning to move away from open source?

deadlyjack:
I don't know how to update there. I have no idea. If anyone can help.

jstetten:
Since ACode is the premiere (IMO) open source coding app and central to the flow of a great deal of open-source-minded developers, I'd like to ask for @IzzySoft's guidance, when he has time — as he's the font of all Android open source repo knowledge.

IzzySoft:
I can take a look (which would have been easier if I had known the package name :wink:). Seems like F-Droid is no longer able to build the app:

* disable: webpack fails
* disable: Task 'assembleRelease' not found in project ':app'.

1.4.164 was the last version with a successful build. 1.5.4 failed with the above errors, 1.5.5 was not even tried as at the tag it still identifies as 1.5.4 (i.e. version was only adjusted AFTER tagging, should have been BEFORE; so lifting the tag to the corresponding commit would trigger another build).

> he's the font of all Android open source repo knowledge.

:flushed: Oops…

deadlyjack:
Do you know how to fix YAML prebuild section? There are some corrections.

```yaml
prebuild:
      - mkdir -p platforms
      - npm install
      - node ./utils/config.js p paid
      - ./node_modules/.bin/webpack --progress --mode production
      - cordova platform add android # change to android@9
      - node ./utils/loadStyles.js
      - sed -i -e '/signingConfig signingConfigs.release/d' -e '/gms/d' -e '/core-google-shortcuts/d'
        platforms/android/app/build.gradle
```

IzzySoft:
No, sorry – I'm not a developer, and completely unfamiliar with node stuff (so for me it would rather be copy-pasting blindly). I couldn't even test that or, if it throws errors, guess what they mean and how to fix. Maybe you could open a merge request with that at fdroiddata? You would most likely know how to interpret output. Should you need help there then, a team member more familiar with node would certainly be able to help.

BEEDELLROKEJULIANLOCKHART:
...Did http://github.com/deadlyjack/Acode/pull/430 remediate this?

IzzySoft:
[...]

PR #430, opened 2022-06-28
Fix Webpack errors in Linux and other unix-based systems

tfuxu:
This PR should fix Webpack errors when compiling on Linux or other unix-based systems.
If applied, it should also resolve issue #410.

I've tested this PR on a containerized Debian Testing with node.js 14.17.6, so those changes should work on F-Droid build servers too.

Ps. I've added sass module to package.json, because without it, I had 72 errors about .sass files in Webpack on my different machine.

### DIFF
--- a/src/components/commandPallete.js
+++ b/src/components/commandPallete.js
@@ -1,7 +1,7 @@
 import tag from 'html-tag-js';
 import Commands from '../lib/ace/commands';
 import helpers from '../lib/utils/helpers';
-import inputhints from './inputHints';
+import inputhints from './inputhints';
 
 export async function commandPallete() {
   const recentlyUsedCommands = RecentlyUsedCommands();
--- a/package.json
+++ b/package.json
@@ -101,6 +101,7 @@
     "marked": "^4.0.17",
     "mime-types": "^2.1.35",
     "mustache": "^4.2.0",
+    "sass": "^1.53.0",
     "url-parse": "^1.5.10",
     "utf8": "^3.0.0",
     "vanilla-picker": "^2.11.2",
--- a/src/components/WebComponents/components.js
+++ b/src/components/WebComponents/components.js
@@ -1,4 +1,4 @@
 import '@ungap/custom-elements';
-import WcPage from './WcPage';
+import WcPage from './wcPage';
 
 customElements.define('wc-page', WcPage);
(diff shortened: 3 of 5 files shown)


src/components/commandPallete.js
```diff
@@ -1,7 +1,7 @@
 import tag from 'html-tag-js';
 import Commands from '../lib/ace/commands';
 import helpers from '../lib/utils/helpers';
-import inputhints from './inputHints';
+import inputhints from './inputhints';
 
 export async function commandPallete() {
```
deadlyjack, line 6: Filename here is correct.

src/components/commandPallete.js
```diff
@@ -1,7 +1,7 @@
 import tag from 'html-tag-js';
 import Commands from '../lib/ace/commands';
 import helpers from '../lib/utils/helpers';
-import inputhints from './inputHints';
+import inputhints from './inputhints';
```
deadlyjack, line 5: No change needed

deadlyjack:
`yarn start` works successfully with these changes?

src/components/commandPallete.js
tfuxu: This file also requires a change, without it webpack produces this error:
```
ERROR in ./src/components/commandPallete.js 6:0-38
Module not found: Error: Can't resolve './inputHints' in '/home/Acode/src/components'
resolve './inputHints' in '/home/Acode/src/components'
  using description file: /home/Acode/package.json (relative path: ./src/components)
    Field 'browser' doesn't contain a valid alias configuration
    using description file: /home/Acode/package.json (relative path: ./src/components/inputHints)
      no extension
        Field 'browser' doesn't contain a valid alias configuration
        /home/Acode/src/components/inputHints doesn't exist
      .js
        Field 'browser' doesn't contain a valid alias configuration
        /home/Acode/src/components/inputHints.js doesn't exist
      .json
        Field 'browser' doesn't contain a valid alias configuration
        /home/Acode/src/components/inputHints.json doesn't exist
      .wasm
        Field 'browser' doesn't contain a valid alias configuration
        /home/Acode/src/components/inputHints.wasm doesn't exist
      as directory
        /home/Acode/src/components/inputHints doesn't exist
 @ ./src/lib/commands.js 33:0-81 69:4-19
 @ ./src/lib/acode.js 11:0-34 73:17-25 74:15-23
 @ ./src/lib/main.js 56:0-28 193:31-36

webpack 5.73.0 compiled with 1 error in 57474 ms
```

tfuxu:
> `yarn start` works successfully with these changes?

Yes it compiled successfully.

PR #542, opened 2022-10-01
Added Gradle to modelist and Updated JS and Kotlin mode

deepkrg17:


deadlyjack (commented):
Why mode-javascript.js is changed?

deepkrg17:
Added prompt as a support function with alert.

PR #645, opened 2023-02-04
New Theme & Some Customisation, etc

bajrangCoder:
- New VSCode Default Dark theme (theme creadit goes to @legendsabbir
- New VictorMono stylish font
- Customised ace autocompleter box
- ~New feature > capy path of opened dir in fileBrowser~
- Added video player like image dialog in editor

Note: Theme and Autocompleter codes are upload in src/ace , so please check it out....

By @bajrangCoder


src/lib/openFile.js
```diff
@@ -70,11 +70,15 @@ export default async function openFile(file, data = {}) {
...
         return;
+      } else if(/video/i.test(fileInfo.type)) {
+        const blob = new Blob([binData], { type: fileInfo.type });
+        dialogs.box(name, `<video src='${URL.createObjectURL(blob)}' controls></video>`);
+        return;
       }
 
       const confirmation = await dialogs.confirm(strings.info, strings['binary file']);
       if (!confirmation) return;
-    }
```
deadlyjack: You need to clean this extra `else if` on line 81

src/pages/fileBrowser/fileBrowser.include.js
```diff
@@ -204,6 +205,14 @@ function FileBrowserInclude(mode, info, doesOpenLast = true) {
         toast(strings.success);
         return;
       }
+
+      if(action === 'copy-path'){
```
deadlyjack: Actually there is no need for this feature because this will copy absolute path and user can't use it anywhere. I'll update the app to add paths in autocomplete like vscode does very soon.

deadlyjack (changes requested):
Thank you so much for this PR, I am not sure why it says that ext-language_tools is added, but it's already there. And I've have added some comments, if it's fixed I can merge it.

src/pages/fileBrowser/fileBrowser.include.js
bajrangCoder: Oo , that's nice

I had added it because many times, user is not able to get the path of dir

bajrangCoder:
As you had said that instead of changing in www/ js/ace directory put that change file in src/ace, that's why added here

src/lib/openFile.js
bajrangCoder: Ok done

deadlyjack:
yes, but it seems to compiled code. what are changes in ext-language_tools?

bajrangCoder:
> yes, but it seems to be compiled code. what are changes in ext-language_tools? See other files in src/ace for example

Just changed background color, hover color, text color of autocompleter box

deadlyjack:
> > yes, but it seems to be compiled code. what are changes in ext-language_tools? See other files in src/ace for example
> 
> Just changed background color, hover color, text color of autocompleter box

Check out this file. You can do the above changes in this scss file and the ext-language_tools

bajrangCoder:
Ok done! Now check .....

PR #659, opened 2023-03-07
Fixed github autobuild action (with sign)

RaSan147:
Using https://github.com/deadlyjack/Acode/pull/592 as base, made some changes and fixes.

Well I'm using a demo sign file, but in your case, you could base64 the sign files and load (and save as files) on run time to sign the apps

Sadly, the pro signed file also stays in the app_dev.zip in github artifact and the .aab file stays in the app_pro.zip

Also tried to fix ftp load fail on last update (but it seems you've removed that)

*also added goto and some of my fav buttons on bottom bar 😅


alexbodn:
hello friend,

thanks for your contribution.
with your action file and a little research,
i could build cordova and the apk files.
however, i have no sign.sh.
i assume it's some private thing, just asking how to make it.

thanks in advance,
alex

RaSan147:
> hello friend,
> 
> thanks for your contribution. with your action file and a little research, i could build cordova and the apk files. however, i have no sign.sh. i assume it's some private thing, just asking how to make it.
> 
> thanks in advance, alex

ah my apologies, this isn't the active auto build version, and sign.sh is open... let me re-post the auto build

PR #691, opened 2023-04-30
Update of French translation

Frigory33:
Hello, thank you for providing Acode! Here is an update of the French translation of the program.

The texts in English look not perfect to me. I suggest you ask somebody to improve them, because it can help translators provide better translations.


deadlyjack:
Thanks for fixing typos and updating translation. I really appreciate it..

Frigory33:
Thank you very much for merging so quickly.

Can you tell me what the `text-transform: capitalize` were for? In French, this style looks very bad. Maybe you only needed the first letter to be capitalized? In that case, you must use the `::first-letter` selector.

I haven’t tested my translations; I hope the use of no-break spaces won’t be a problem.

PR #703, opened 2023-05-24
Add fastlane metadata for F-Droid

### LINKED ISSUE #183: Screenshots for F-Droid
You can add Screenshots for F-Droid, too, if you want.
https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/
https://gitlab.com/snippets/1895688
That was simply information about what is still possible. You would then only have to enter the
screenshots in fastlane. This is just one way that you can also add screenshots to the app view
if you can. If you added screenshots (then F-Droid is happy) and if you can't add any -> then
you can close the issue because it was only an info. Thank you that your app is available in
fdroid. Very much for your attention.

shuvashish76:
Closes #183


deadlyjack:
Thanks for this. I appreciate it.

PR #941, opened 2024-03-10
Termux integration

7HR4IZ3:

## Example Usage

```jsx

// Create new connection
  command: "bash", args: [],
  onmessage(stdout) {
    console.log(stdout);
  }
});

// Add listeners

// Kill connection
connection.kll();
```

# Type Interface

## Command

## **PtyResponse**

- `stderr`: String
- `stdout`: String
- `exitCode`: Number
- `stderrLength`: Number
- `stdoutLength`: Number

## PtyError

- `error`: String
- `errorCode`: Number
- `exitCode`: Number

## RunCommand


deadlyjack:
Hi, thanks for this PR. Seems to be an amazing update. As there are too many changes, I'll have to review it carefully and merge it.

7HR4IZ3:
Alright.. I also had to make some modification to make it run on my device.

- Like changing the package to the free version.
- Adding tailwindcss (Pushed it by mistake)
- Adding notifications plugin to notify users when a command is running
- Adding http plugin to bypass cors limit on browser fetch function (planning on adding websocket server to the server plugin so it can be used to add live reload functions to the browser)

Also we still have to figure out a way to prevent plugins from running commands without the user's permission, like adding:

```json
...
"permissions": ["acode.pty.command"]
...
```

But then we'll need to figure out a way to identify which plugin is running the command..

deadlyjack:
> Also we still have to figure out a way to prevent plugins from running commands without the user's permission, like adding:

Yeah, this is must. Let me get some free time and add the permission mechanism in the Acode.

ByteJoseph (approved):

PR #984, opened 2024-06-27
feat: Added support for spawing termux from acode using pty module

7HR4IZ3:

## Example Usage

```jsx

// Create new connection
  command: "bash", args: [],
  onmessage(stdout) {
    console.log(stdout);
  }
});

// Add listeners

// Kill connection
connection.kll();
```

# Type Interface

## Command

## **PtyResponse**

- `stderr`: String
- `stdout`: String
- `exitCode`: Number
- `stderrLength`: Number
- `stdoutLength`: Number

## PtyError

- `error`: String
- `errorCode`: Number
- `exitCode`: Number

## RunCommand


bajrangCoder:
Seems fine , let me try ...

7HR4IZ3:
I'm planning on adding a permissions module too which stores pluginID and the modules they have access to. To check permission, the module would call a `permissions.checkPermission(permissionName, plugin)` function). It would show a simple dialog whenever a plugin tries to use the plugin to confirm the user allows the plugin to use the module.

So since we can't just pass the plugin id as a string, lets add a third argument to the `acode.setPluninInit` function, a class instance of `AcodePlugin`, it would hold all the items should to be in the original third argument but will also store the values of the `plugin.json` file for that plugin, so a plugin cannot pretend to be another plugin.

_Example_
```javascript
// plugin.js
acode.setPluginInit(plugin.id, (baseUrl, $page, plugin) => {
    const {
        name, id, cacheFile, cacheFileUrl,
        isJustInstalled
    } = plugin;
    const pty = acode.require("pty");
    pty.host.run({
        command: "ls", args: [],
        onmessage(stdout) {
            console.log(stdout);
        }
    });


    // To request permission
    const permissions = acode.require("permissions");
    const hasPermission = permissions.check("pty.RUN_COMMAND", plugin);
    if (!hasPermission) {
       permissions.request("pty.RUN_COMMAND", plugin);
    }
    else { ... }
});

// pty.js (pty.host.run function)
run({ ..., plugin }) {
    if (!(plugin instanceof AcodePlugin)) throw new Error("...");

    const permissions = acode.require("permissions");
    const hasPermission = permissions.check("pty.RUN_COMMAND", plugin);
    if (!hasPermission) throw new Error("...")
    else { ... }
}

```

bajrangCoder:
Good idea but not a good approach that will break a lots of plug-in 

I think permission should be only for critical modules like pty, No need in normal module

7HR4IZ3:
It shouldn't break old plugins since the third argument for `acode.setPluginInit` is usually `{ cacheFile, cacheFileUrl }`, which would still work with old plugins. And the permissions module can be plugged in to new modules that require it, like the `pty` module.

7HR4IZ3:
@bajrangCoder Were you able to test this?

bajrangCoder:
its not working!
Even termux session is running background, there is no response from termux

I'm on Android 14

7HR4IZ3:
Fixed. I forgot to register the service.
Add this to the `platform` section in `config.xml`:

```xml
<config-file target="AndroidManifest.xml" parent="./application">
    <service android:name="com.foxdebug.system.PluginResultService" />
</config-file>
```

Tested on Android 13 (will download android 14 soon).

Screenshot 2024-07-25 at 2 51 12 PM


Also regarding termux running in the background:

In the `system` global variable, I added a `run` function to interact with termux (usually to start `acode-pty`) so you can run this:

```javascript
system.run({
    command: "/data/data/com.termux/files/usr/bin/ls",
    workDir: "/data/data/com.termux/files/home", args: ["-a"],
    background: false, sessionAction: 3
  }).then(console.log)
```

If you set `background` to `true`, it would run the code in the background without waiting for termux, but if it's `false` then you have to open termux after running the code but you can customize the behaviour using the `sessionAction` option  (https://github.com/termux/termux-app/wiki/RUN_COMMAND-Intent#run_command-intent-command-extras)

bajrangCoder:
You have messed up your changes while merging the main branch.

On your project branch there is no any system.run methods

7HR4IZ3:
I'll resubmit the pull request again on another branch 🥲

PR #997, opened 2024-07-26
Termux integration

7HR4IZ3:

### System Run
```javascript
system.run({
    command: "/data/data/com.termux/files/usr/bin/ls",
    workDir: "/data/data/com.termux/files/home", args: ["-a"],
    background: false, sessionAction: 3
  }).then(console.log)
```

Read this ([TERMUX.RUN_COMMAND](https://github.com/termux/termux-app/wiki/RUN_COMMAND-Intent#run_command-intent-command-extras)) for more info on `sessionAction` and other arguments 

### Using Acode PTY

You should use this for spawning and interacting with processes

```javascript

// Create new connection
  command: "bash", args: [],
  onmessage(stdout) {
    console.log(stdout);
  }
});

// Add listeners

// Kill connection
connection.kll();
```

# Type Interface

## Command

## PtyResponse

- `stderr`: String
- `stdout`: String
- `exitCode`: Number
- `stderrLength`: Number
- `stdoutLength`: Number

## PtyError

- `error`: String
- `errorCode`: Number
- `exitCode`: Number

## RunCommand

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -38,6 +38,9 @@
             <application android:hardwareAccelerated="true" />
             <application android:requestLegacyExternalStorage="true" />
             <application android:largeHeap="true" />
+            <application>
+                <service android:name="com.foxdebug.system.PluginResultService" />
+            </application>
         </edit-config>
 
         <config-file parent="./application/activity" target="AndroidManifest.xml">
@@ -68,11 +71,15 @@
         </config-file>
 
         <config-file target="AndroidManifest.xml" parent="/manifest">
+            <queries>
+                <package android:name="com.termux" />
+            </queries>
+
             <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
             <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
             <uses-permission android:name="android.permission.WRITE_MEDIA_STORAGE" />
             <uses-permission android:name="android.permission.VIBRATE" />
-            <!-- <uses-permission android:name="com.termux.permission.RUN_COMMAND" /> -->
+            <uses-permission android:name="com.termux.permission.RUN_COMMAND" />
         </config-file>
 
         <hook type="before_prepare" src="hooks/modify-java-files.js" />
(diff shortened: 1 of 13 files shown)


bajrangCoder:
> @bajrangCoder Could you test this out?

I will check it , I will notify if any changes will need

config.xml
bajrangCoder, line 1: Is it necessary to specify the gradle and build tool version

7HR4IZ3: No. That's just for my setup

bajrangCoder: Ok

bajrangCoder:
I have tried many times but haven't succeeded, even after making modifications. None of my attempts have been successful; almost every time I try, it fails. Since this solution isn't working, I can't merge it. It needs to work flawlessly on supported devices. Please refine your implementation and also share a test APK for testing.

Additionally, why is there a need to ask for permission to run a command in Termux if the user doesn't have Termux installed?

I think your pull request is quite tangled. Have you tried building the APK from this PR? Did you succeed?

RohitKushvaha01:
> I have tried many times but haven't succeeded, even after making modifications. None of my attempts have been successful; almost every time I try, it fails. Since this solution isn't working, I can't merge it. It needs to work flawlessly on supported devices. Please refine your implementation and also share a test APK for testing.
> 
> Additionally, why is there a need to ask for permission to run a command in Termux if the user doesn't have Termux installed?
> 
> I think your pull request is quite tangled. Have you tried building the APK from this PR? Did you succeed?

Issues with TERMUX.RUN_COMMAND:

Requires allow-external-apps=true in ~/.termux/termux.properties.

Needs execution permission in settings; some devices don’t show the option, forcing users to use ADB.

Termux get frozen in the background, preventing them from receiving the intent. There's no way to confirm if the command ran (PendingIntent could help).

On Android 14+, the F-Droid version of Termux can't be installed on some devices. The Play Store version doesn't support run_command.

I worked around the freeze issue in Xed-Editor by launching Termux before sending the intent, but this pops open the app even when running commands silently potentially confusing the user.

PR #1020, opened 2024-08-29
Implement Logger system

bajrangCoder:
- Logger system
- Removed some old `console.log` and `console.error` with logger api
- Added new log method to `window` object
- Now avoid using console logs instead use logger , so it will be easy to debug any bug
- Logging system is efficient and it automatically manages flushing of buffer and LRU algorithm to keep log file small and relevant 
- Search for "Open Log file" in command palette to open the log file in editor.

### DIFF
--- a/src/pages/plugin/plugin.js
+++ b/src/pages/plugin/plugin.js
@@ -140,7 +140,7 @@ export default async function PluginInclude(
 					}
 				}
 			} catch (error) {
-				console.error(error);
+				window.log("error", error);
 			} finally {
 				loader.removeTitleLoader();
 			}
@@ -154,6 +154,7 @@ export default async function PluginInclude(
 			$button?.click();
 		}
 	} catch (err) {
+		window.log("error", err);
 		helpers.error(err);
 	} finally {
 		loader.removeTitleLoader();
@@ -173,6 +174,7 @@ export default async function PluginInclude(
 			}
 			render();
 		} catch (err) {
+			window.log("error", err);
 			helpers.error(err);
 		}
 	}
@@ -190,6 +192,7 @@ export default async function PluginInclude(
 			}
 			render();
 		} catch (err) {
+			window.log("error", err);
 			helpers.error(err);
 		}
 	}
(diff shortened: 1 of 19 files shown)


src/pages/plugin/plugin.js
```diff
@@ -231,6 +234,8 @@ export default async function PluginInclude(
 				$button.textContent = oldText;
 			}
 		} catch (error) {
+			window.log("error", "Failed to buy:");
```
UnschooledGamer, line 237: @bajrangCoder I think you forgot to append the caught `error` to the log message

src/pages/plugins/plugins.js
```diff
@@ -253,7 +254,8 @@ export default function PluginsInclude(updates) {
       return await response.json();
     } catch (error) {
       $list.all.setAttribute("empty-msg", strings["error"]);
-      console.error(error);
+      window.log("error", "Failed to filter plugins:");
+      window.log("error", error);
```
UnschooledGamer, line 258: Used to do one Job Twice

src/sidebarApps/extensions/index.js
```diff
@@ -157,7 +157,8 @@ async function filterPlugins() {
 			updateHeight($searchResult);
 		}
 	} catch (error) {
-		console.error("Error filtering plugins:", error);
+		window.log("error", "Error filtering plugins:");
+		window.log("error", error);
```
UnschooledGamer, line 161: 👀

UnschooledGamer (changes requested):

UnschooledGamer:
By-mistake Selected the Wrong Review Type. Sorry.

src/pages/plugin/plugin.js
bajrangCoder: No , check second line.
It's like that because if I will pass `Error` object combined with the the string then the `Error` object will become string and then it will not show the full error

src/pages/plugins/plugins.js
bajrangCoder: In first time sending the string log then sending the actual error object

PR #1030, opened 2024-09-06
Plugin Dependencies

alMukaafih:


src/lib/installPlugin.js
```diff
@@ -139,3 +155,99 @@ async function createFileRecursive(parent, dir) {
...
+		)
+			.readFile("json")
+			.catch(() => null);
+
+		if (!remoteDependency) return true;
+
+		const version = await isInstalled(id);
+		if (remoteDependency?.version === version) return false;
+
+		plugin = Object.assign({}, remoteDependency);
```
bajrangCoder: i guess you forget to define this variable

bajrangCoder:
have you tested it? It doesn't work

alMukaafih:
I have tested it and it works for  both free and paid plugins

src/lib/installPlugin.js
alMukaafih: The variable is unused

bajrangCoder:
> I have tested it and it works for both free and paid plugins though further testing is needed for paid plugins

Now it's working fine.
Just one thing left:
It should show the name of plugin which is installing in loader and also before installing the dependencies it's nice to show the user that it's going to install these plugins too.

alMukaafih:
> > I have tested it and it works for both free and paid plugins though further testing is needed for paid plugins
> 
> Now it's working fine. Just one thing left: It should show the name of plugin which is installing in loader and also before installing the dependencies it's nice to show the user that it's going to install these plugins too.

Sorry for the late response, was facing some issues irl.

I am deliberately either to show an `alert` for the dependencies that are going to be installed or a `confirmation`.
Something like
```js
alert(
    "Installer Notice",
    "Acode is going to install the following dependencies<br/>"
      + plugin.dependencies.join("<br/>"),
    () => {
        // Resolve dependencies
    },
);
```
Or
```js
const shouldInstall = await confirm(
    "Installer Notice",
    "Acode is going to install the following dependencies<br/>"
      + plugin.dependencies.join("<br/>"),
);

if (shouldInstall) {
    // Resolve dependencies 
}

bajrangCoder:
Use confirmation

alMukaafih:
> Use confirmation

Ok 👍

bajrangCoder (approved):

Reybabylon (approved):

Reybabylon (approved):

Issue #1026, opened 2024-09-04
Enhancements to Acode Plugin installer 


alMukaafih:
Is it possible to add support for plugin dependencies?

bajrangCoder:
Clearly explain it in detail

alMukaafih:
Like an array of plugins in `plugin.json` to be Installed along with the plugin. Maybe something like
```json
{
  "dependencies": [
    "bajrangcoder.acodex", "sebastianjnuwu.acode.eruda"
  ]
}
```

alMukaafih:
I am willing to assist in implementing this

bajrangCoder:
> Like an array of plugins in `plugin.json` to be Installed along with the plugin. Maybe something like
> 
> ```json
> {
>   "dependencies": [
>     "bajrangcoder.acodex", "sebastianjnuwu.acode.eruda"
>   ]
> }
> ```

Ok

bajrangCoder:
> I am willing to assist in implementing this

Contributions are welcomed ☺️, feel free to open pr

sebastianjnuwu:
> > I am willing to assist in implementing this
> 
> Contributions are welcomed ☺️, feel free to open pr

I had already talked about this, the system is similar to that of vscode, the .acode folder has the settings and dependencies and acode recognizes itself when the project folder is open if not with the default

alMukaafih:
Currently when updating Plugins the installer does not delete the previous installation of the plugin rather it installs the update over the current installation. There is a problem of redundant files.
Suppose this is the current installation
```
.
├── assets
│   ├── 0.svg
│   ├── 1.svg
│   └── 2.svg
├── icon.png
├── main.js
├── plugin.json
└── readme.md
```
and this is the update
```
.
├── assets
│   ├── 3.svg
│   ├── 4.svg
│   └── 5.svg
├── icon.png
├── main.js
├── plugin.json
└── readme.md
```
after updating this will be the current installation
```
.
├── assets
│   ├── 0.svg
│   ├── 1.svg
│   ├── 2.svg
│   ├── 3.svg
│   ├── 4.svg
│   └── 5.svg
├── icon.png
├── main.js
├── plugin.json
└── readme.md
```

bajrangCoder:
Nice, I haven't noticed this.

> I suggest using a combination of id and version as the plugin folder name

I think it will be good to remove the previous folder and then add the new update one , as the the id and version combination will make different instances of same plugin.

If you have any better idea then please share

[...]

PR #1058, opened 2024-10-20
Added install state for better update of plugins (#1026)

alMukaafih:
Fix for #1026


bajrangCoder (approved):

bajrangCoder:
Theoretically it seems correct, have you tested?

bajrangCoder:
I found a bug. When installing some plugins locally it fails with no any error logs , just shows a toast.
You can reproduce it by building this plugin and try it by installing locally.

https://github.com/bajrangCoder/MarkdownCompiler

PR #1072, opened 2024-11-12
Update hu-hu.json

summoner001:
Translate new strings.


bajrangCoder:
Thanks for these quick translations ♥️

PR #1085, opened 2024-11-24
fix: Acode ignoring main, readme and icon fields in plugin manifest

alMukaafih:
Fix to Acode Ignoring main, readme and icon fields in plugin manifest (#1026)


bajrangCoder:
It fails to load existing plugins

alMukaafih:
> It fails to load existing plugins

I will look into it

bajrangCoder:
I think problem is in plugin templates , as it points to `main: "dist/main.js"` but in zip file it's not in dist folder

alMukaafih:
> I think problem is in plugin templates , as it points to `main: "dist/main.js"` but in zip file it's not in dist folder .
> 
> So handle this as legacy one

I noticed this issue too. That is why I was testing for the existence of `pluginJson.main` before defaulting to `"main.js"` but it seems something went wrong somewhere.

bajrangCoder:
The issue persists when loading previously published plugins; we need to implement a legacy solution to ensure compatibility.

alMukaafih:
> The issue persists when loading previously published plugins; we need to implement a legacy solution to ensure compatibility.

I tried it on a clean install and it works. On an update it fails to load previously installed plugins.

bajrangCoder:
To maintain compatibility, you can implement a simple check during the plugin load process. If the `main` field is set to `dist/main.js`, treat it as `main.js` instead. That's the main cause for problems.

alMukaafih:
> To maintain compatibility, you can implement a simple check during the plugin load process. If the `main` field is set to `dist/main.js`, treat it as `main.js` instead. That's the main cause for problems.

I have implemented a more comprehensive check with my latest push. If the file that the `main` field points to does not exist, it falls back to the previous implementation of loading `main.js` .

bajrangCoder:
Oo that's nice . But there are errors when listing local files in sidebar (I have tried your latest commit)

ByteJoseph (commented):
👀

bajrangCoder:
@alMukaafih 
It's ready to merge, but I've noticed some spikes in app opening times, even with just 6-7 plugins. I'm not sure if this is related to this change or not.

alMukaafih:
It is most likely due to extra fs call (test for plugin.main) while loading the plugins. I will reimplement this test using try/catch

alMukaafih:
I have reimplemented the test using try/catch but there in still an unavoidable call to load the `plugin.json` for each plugin just to get the main field. We can actually reduce the call to just one by caching the main field of each plugin. The cache can be something like this
```
[{ "id": "acode.plugin", "main": "main.js" }]
```
This will reduce the number of fs calls while loading plugins.

We could also defer the loading of plugins to after the Acode has fully opened but could potentially break some plugins.

bajrangCoder:
> I have reimplemented the test using try/catch but there in still an unavoidable call to load the plugin.json for each plugin just to get the main field.

Ok !
 I'll test it

bajrangCoder:
> We could also defer the loading of plugins to after the Acode has fully opened so that Acode can open faster....

Yes, it's a critical problem

bajrangCoder:
> We can actually reduce the call to just one by caching the main field of each plugin. The cache can be something like this.....

No need

bajrangCoder:
> > I have reimplemented the test using try/catch but there in still an unavoidable call to load the plugin.json for each plugin just to get the main field.
> 
> Ok ! I'll test it

It's working, but there is a issue: As when Acode failed to load plugins then it deletes that plugin.

alMukaafih:
> > > I have reimplemented the test using try/catch but there in still an unavoidable call to load the plugin.json for each plugin just to get the main field.
> > 
> > 
> > Ok ! I'll test it
> 
> It's working, but there is a issue: As when Acode failed to load plugins then it deletes that plugin.

Is it while Acode is opening or when installing a plugin?

bajrangCoder:
> > > > I have reimplemented the test using try/catch but there in still an unavoidable call to load the plugin.json for each plugin just to get the main field.
> > > 
> > > 
> > > Ok ! I'll test it
> > 
> > 
> > It's working, but there is a issue: As when Acode failed to load plugins then it deletes that plugin.
> 
> Is it while Acode is opening or when installing a plugin?

Opening the Acode (for existing plugins which aren't patched)

noob-code55 (approved):

PR #1138, opened 2025-01-07
Switch syntax for testing file presence

eclipseo:
[[ ]] is a bashism, adopted by some other shells (zsh, ksh) but not all. It does provide extra features (pattern matching and regex).

The POSIX way would be to use `test` or it's synonym ` [  ] `.  Both forms are equivalent, and the square brackets version [ ] requires spaces around the brackets as well as around each operator and operand. Here are some examples:

 ‐ `if [ -f "file.txt" ]; then echo "File exists"; fi`
 - `if test -f "file.txt"; then echo "File exists"; fi`

The POSIX standard does not include the double brackets [[ ]], so for scripts that need to be portable across different POSIX-compliant shells, we should stick with [ ] or test.

### DIFF
--- a/src/fileSystem/sftp.js
+++ b/src/fileSystem/sftp.js
@@ -124,7 +124,7 @@ class SftpClient {
 					}
 
 					const file = this.#safeName(filename);
-					const cmd = `[[ -f "${file}" ]] && echo "Already exists" || touch "${filename}"`;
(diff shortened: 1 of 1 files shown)


eclipseo:
Note that I donx̌t have the means to test this, ai don't have a build environment for Android and so on.

bajrangCoder:
This isn't the solution, instead we can use `test -f filename`

bajrangCoder:
#1167 will fix this issue

PR #1210, opened 2025-03-16
feat: Add file android.yml

iKiwo:
Add the android.yml file so you can perform actions and compile the project in case anyone needs it.

You can modify it if necessary.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,0 +1,34 @@
+name: Android Build
+
+on:
+  push:
+    paths:
+      - '.github/workflows/android.yml'
+  workflow_dispatch:
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


bajrangCoder:
It will be good to trigger it on pr only (and the build apk should be only for testing, not for regular uses)

UnschooledGamer:
Current workflow code, Is just for manual Trigger of workflow or when the workflow has any changes received via push event. 


So it's just a one time apk build workflow such as for releases etc.

bajrangCoder:
> Current workflow code, Is just for manual Trigger of workflow or when the workflow file has any changes received via push event.
> 
> So it's just a one time apk build workflow such as for releases etc.

Don't need these kind of workflows

PR #1244, opened 2025-04-07
feat: inapp acode account login

bajrangCoder:
### Demo

Screenshot_20250407-172107

Screenshot_20250407-202348~2

### DIFF
--- a/src/components/sidebar/index.js
+++ b/src/components/sidebar/index.js
@@ -1,6 +1,8 @@
+import toast from "components/toast";
 import "./style.scss";
 import Ref from "html-tag-js/ref";
 import actionStack from "lib/actionStack";
+import auth, { loginEvents } from "lib/auth";
 import constants from "lib/constants";
 
 let $sidebar;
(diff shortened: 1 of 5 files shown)


bajrangCoder:
> Before merging I would suggest not displaying email address by default so that we ourselves don't have to spend extra time editing screenshots or screen recordings just to hide it

That's a public email. 
Btw why anybody will take screenshot of sidebar with opened menu?

deadlyjack (approved):

deadlyjack:
You're right... we can mask the email. And use user id instead in the url.

bajrangCoder:
> > [*](https://github.com/Acode-Foundation/Acode/pull/1244#issuecomment-2785095150) That's a public email.
> 
> Some people may want to prefer to use their private email to login with, which for those with privacy concerns is not recommended when the site also publicly displays the email*, although if they used their private email already they should change it to a public email in the account settings for now
> 
> *In acode website's case the email it's just visible on the profile's URL Screenshot_20250409-033036_Chrome

That's a issue on website but there is no problem with showing email in app(like a profile or something)

Issue #1091, opened 2024-12-04
Add support for relative assets handling in built in server 


Inori-dokusha:
### Check for existing issues

- [x] Completed

### Describe the bug / provide steps to reproduce it

![Screenshot_2024-12-04-10-55-49-01_a2cf8efcdd42a8e6f7906303f104fb67](https://github.com/user-attachments/assets/7de04399-4c13-419d-95c1-b2943b9e8220)

When the html file is liveserver, it will not be connected to the CSS in a different folder but in one folder as its parent.

### Environment

- Android 11
- Acode 1.10.5
- Webview 132.0.6834.14

### If applicable, attach your Acode.log file to this issue.

<details><summary>Acode.log</summary><pre>

</pre></details>

Inori-dokusha:
Please fix this part bro, because it's really important

Issue #651, opened 2023-02-14
Markdown image preview not working when images are in a different folder


AlbertoFabbri93:
Folder structure:
```
│
└───notes_folder
│        note1.md
│        note2.md
│   
└───images_folder
           img1.svg
           img2.svg    
```
If I reference `img1.svg` or `img2.svg` from one of the markdown notes Acode does not display them. I am including the images using the html `<img>` tag.

sebastianjnuwu:
Still having the same problem?

deadlyjack:
You can only reference image from same directory or in child directory but not parent's sibling directory.

AlbertoFabbri93:
@deadlyjack thank you for the reply. It would be nice to have this feature, vs code, for example, is able to display the images available in the parent's sibling directories.

AlbertoFabbri93:
Can you tell me in what version of Acode is this fixed? Because it doesn't work for me with the current version.

PR #1257, opened 2025-04-12
fix. http relative path

RohitKushvaha01:
Closes: #1091 , #651

HTML viewer now resolves relative paths

Before:
image

Now:
image

This change might break something. I recommend testing it before merging.


bajrangCoder:
There are multiple problems:

- saf uri contains many special stuffs like `::primary:`, or termux saf uri etc which needs to be stripped out 
- also problem with SFTP/ftp url as they typically starts with protocol SFTP/ftp 
- and there should be fallback if there is no folders in added folder or any other case then just serve the current document

RohitKushvaha01:
- [x] Internal Storage
- [x] Termux
- [ ] FTP
- [ ] Redirect to current file

bajrangCoder:
Now things are fine. But still needs some fixes in these cases:

- when a file is opened which is not saved
- when a file is opened and there is no folders in sidebar 
- when a file is opened but that's not the part of any Opened folder in sidebar

In these cases it should fallback as previous.

When running a `.js` file it opens browser but it should open js console

RohitKushvaha01:
@bajrangCoder can you test this?, It's working in my device

bajrangCoder:
> @bajrangCoder can you test this?, It's working in my device

I'll test it

PR #1259, opened 2025-04-13
feat: rainbow bracket pair colorizer

bajrangCoder:
This pr adds Rainbow brackets pair colorizer similar to vscode one.
- it is fast
- it is easily toggle able from editor settings
- can work in any files

### Demo

Screenshot_20250413-223130


bajrangCoder:
> So, we should implement it ourselves as a plugin?

There is a plugin available(`Extra Syntax Highlights`)
This pr approach is fast , efficient but not good for usabilty due to multiple token/dom refresh, as it uses post tokenization

PR #1286, opened 2025-04-27
Update zh-cn.json and zh-hant.json

LaunchLee:


bajrangCoder (commented):
I don't know this language but google translator translate it to : `"Gradient Fold Component"`

Is it correct?

LaunchLee:
The "fade" is translated to "淡入淡出" in simplified Chinese (These characters is the same as traditional Chinese) which says "fade in and fade out". I was just tried to think a shorter word with the similar meaning that resulted in "渐变式" but I did not notice that it can be referred to the word "gradient(渐变) form(式)" which is saying color, thought it reduces only one character. For common sense, it's better to be translated to "淡入淡出".

LaunchLee:
En...In my opinion, the both translations are no problem. "渐变式" is more general, "淡入淡出" is more specific.

LaunchLee:
What is the "widget" referred to in Acode interface? Honestly speaking I do not find which part does this config change in Acode interface.

bajrangCoder:
> What is the "widgets" referred to in Acode interface?

It's arrow like thing available in editor gutter to fold codes.

This `小部件` is perfect for widget.

LaunchLee:
> It's arrow like thing available in editor gutter to fold codes.

OK, now I have gotten to know it. Widgets in android home screen are translated to ```小部件```. Treating these arrows as widget is a little exaggerating. At first it is difficult to think the ```fold widgets``` as the ```code-folding buttons``` in my mind. From the function perspective, ```淡入淡出代码折叠按钮``` is more intuitive.

zh-cn.json:
``` json
"Fade fold widgets": "淡入淡出代码折叠按钮"
```

zh-hant.json:
``` json
"Fade fold widgets": "淡入淡出代碼折疊按鈕"
```

bajrangCoder:
> Treating these arrows as widget is a little exaggerating.

Widget has broader meaning so it isn't to be confused or limited. Also the ace library which we are using in Acode call them widgets , not a button

LaunchLee:
Oh……I see. My native language is simplified Chinese, so I think it in Chinese language style. My fault. I had a look in [Wikipedia](https://zh.wikipedia.org/wiki/Widget):

> [控件](https://zh.wikipedia.org/wiki/%E6%8E%A7%E4%BB%B6)（英语：widget），图形使用者界面（GUI）中的互动元素（Interactive Elements），例如按钮（Buttons）或卷轴（Scrollbars）

So finally, I decide to take this translation as I have translated yet:

zh-cn.json:
``` json
"Fade fold widgets": "淡入淡出代码折叠按钮"
```

zh-hant.json:
``` json
"Fade fold widgets": "淡入淡出代碼折疊按鈕"
```

PR #1317, opened 2025-05-20
Fix html content access

RohitKushvaha01:
**Note:** If your HTML file is not inside an opened project folder, it can only access files in the same folder or below it  not outside.

### For example:

* Let's say your HTML file is at:
  `/sdcard/HelloWorld/html/index.html`

---

✅ **Case 1: You open `/sdcard/HelloWorld` as the project folder**

* Now the HTML file **can access**:

  * `/sdcard/HelloWorld/css/style.css`
  * `/sdcard/HelloWorld/images/logo.png`
  * `/sdcard/HelloWorld/html/scripts/app.js`

* But it **cannot access**:

  * `/sdcard/AnotherFolder/data.json`
    (because it's outside the project root)

---

❌ **Case 2: No folder is opened in the sidebar or a different folder is opened (just the HTML file)**

* Now the root becomes `/sdcard/HelloWorld/html`, so the HTML file **can only access**:

  * `/sdcard/HelloWorld/html/scripts/app.js`
  * `/sdcard/HelloWorld/html/images/logo.png`

* It **cannot access**:

  * `/sdcard/HelloWorld/css/style.css`
    (because it’s outside `/sdcard/HelloWorld/html`)

### DIFF
--- a/src/lib/run.js
+++ b/src/lib/run.js
@@ -180,14 +180,14 @@ async function run(
 		const reqId = req.requestId;
 		let reqPath = req.path.substring(1);
 
-		console.log(`XREQPATH ${reqPath}`)
-		console.log(req)
+		console.log(`XREQPATH ${reqPath}`);
+		console.log(req);
 
 		if (!reqPath || (reqPath.endsWith("/") && reqPath.length === 1)) {
 			reqPath = getRelativePath();
 		}
 
-		console.log(`XREQPATH1 ${reqPath}`)
+		console.log(`XREQPATH1 ${reqPath}`);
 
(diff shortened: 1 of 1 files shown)


UnschooledGamer:
> It cannot access:

/sdcard/HelloWorld/css/style.css
(because it’s outside /sdcard/HelloWorld/html)

For case 2, It's about detecting the parent directory of the file(opened the html file) and limiting the access to it?

UnschooledGamer (approved):
Seems ok

bajrangCoder (approved):

PR #1318, opened 2025-05-21
fix. path overlap issue in html viewer

RohitKushvaha01:


RohitKushvaha01:
@bajrangCoder, this should fix the infinite loading issue.

bajrangCoder (approved):

PR #1338, opened 2025-06-02
Added Executor

RohitKushvaha01:
#### Example usage:

```javascript
Executor.exec(
    "echo hello",
    (stdout) => {
        console.log("Executor output: " + stdout);
    },
    (error) => {
        console.warn("Executor error: " + error);
    }
);
```

---

### ⚠️ Limitations

1. **No W^X support**:
   Starting from Android 9, executing binary files is not allowed

   * ✅ **Workaround**: You can use `/system/bin/linker` or `/system/bin/linker64` to load and execute non-static binaries, but **this only works if the binary is outside internal storage** and is **dynamically linked**.

2. **Shells are not fully supported**:
   You can execute shell scripts using shell interpreters (like `sh` or `bash`), but **do not execute the shell directly** (e.g., just `"sh"`), as it will hang. shells remain active and wait for input until explicitly terminated via `Ctrl+C` or a kill signal.

### DIFF
--- a/src/plugins/Executor/www/Executor.js
+++ b/src/plugins/Executor/www/Executor.js
@@ -1,5 +1,0 @@
-var exec = require('cordova/exec');
-
-exports.coolMethod = function (arg0, success, error) {
-    exec(success, error, 'Executor', 'coolMethod', [arg0]);
-};


bajrangCoder (approved):

syed-rasel:
Hi hes

Issue #878, opened 2023-11-01
Repositioning tab also scrolls the tab bar... (1.8.7/330)


ghost:
...regardless of whether it's near the egde of the screen or not
https://youtu.be/3QG52RFpN94

deadlyjack:
Seems to be working fine for me. Can you please tell your android version and device manufacturer.

1Code-0:
![Screenshot_20230823-082722_Settings](https://github.com/deadlyjack/Acode/assets/113766442/08d30368-d0e7-48d6-9bd3-691a3f366975)
![Screenshot_20230823-082949_Settings](https://github.com/deadlyjack/Acode/assets/113766442/f5567dba-9bd5-4b62-a615-133a8b854e57)

peasneovoyager2banana2:
Still an issue

https://github.com/user-attachments/assets/55967993-e319-4272-9463-650ec0ce1aaf

UnschooledGamer:
> Still an issue https://github.com/user-attachments/assets/9210ac6a-4d32-4b2b-9ad8-c8b3d81b21a2

In the video, you're hitting the scroll points, which is the reason it's scrolling, so you can reposition the tab further.

peasneovoyager2banana2:
> In the video, you're hitting the scroll points, which is the reason it's scrolling, so you can reposition the tab further.

Please watch the re-recorded video on https://github.com/Acode-Foundation/Acode/issues/878#issuecomment-2975351882, I didn't start from or hit the edges unlike as seen in that YouTube video

PR #1382, opened 2025-06-20
fix: unintentional scrolling of tab container while dragging a tab

peasneovoyager2banana2:
Fixes: https://github.com/Acode-Foundation/Acode/issues/878

### DIFF
--- a/src/handlers/editorFileTab.js
+++ b/src/handlers/editorFileTab.js
@@ -62,6 +62,10 @@ let parentRight = 0;
  * @type {number}
  */
 let animationFrame = null;
+/**
+ * @type {number}
+ */
+let prevScrollLeft = 0;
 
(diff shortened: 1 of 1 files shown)


peasneovoyager2banana2:
### Before
https://github.com/Acode-Foundation/Acode/issues/878#issuecomment-2975351882
### After
https://github.com/user-attachments/assets/773c80fb-c8e0-4997-84c2-ba04878ec7eb

peasneovoyager2banana2:
> I like to test myself before merging

[Test APK](https://github.com/peasneovoyager2banana2/Acode/actions/runs/15782049738)

Seeing how long it takes and how complicated it is to set up and build locally on Android alone (and also because I don't have a PC), I built the APK using actions instead

PR #1387, opened 2025-06-23
Integrated cordova-plugin-buildinfo as a local plugin

### LINKED ISSUE #1377: Remove some non-determinism
### Check for existing issues

- [x] Completed

Your app is not built reproducible in F-Droid, but we continuously test older versions on https://verification.f-droid.org and would like more and more apps to become repro

Looking at the latest report for your app: https://verification.f-droid.org/packages/com.foxdebug.acode/

Can you track where is that `BUILDINFO_TIMESTAMP` set and maybe set it to be based on the last commit eg. `git log -1 --pretty=%ct` ?

Or better yet, F-Droid implemented `SOURCE_DATE_EPOCH` environment variable, based on the last commit so you could try to use that or, if not set, fallback to the current solution.

ref: https://gitlab.com/fdroid/fdroidserver/-/merge_requests/1653

ref: https://reproducible-builds.org/docs/source-date-epoch/

### Environment

F-Droid build server

_No response_

</pre></details>

RohitKushvaha01:
Closes #1377 

removed _BUILDINFO_TIMESTAMP


bajrangCoder (commented):
Isn't there any other way

UnschooledGamer (approved):
Fine for now!

RohitKushvaha01:
> Isn't there any other way

We could remove the plugin entirely, but other parts of the code might still depend on it. That’s why I cloned it locally and made the changes instead.

UnschooledGamer:
> Or better yet, F-Droid implemented SOURCE_DATE_EPOCH environment variable, based on the last commit so you could try to use that or, if not set, fallback to the current solution.

Could do the above ^^

OR fallback commit date or current date as last option.

I wonder why it started failing recently 🤔

RohitKushvaha01:
> > Or better yet, F-Droid implemented SOURCE_DATE_EPOCH environment variable, based on the last commit so you could try to use that or, if not set, fallback to the current solution.
> 
> Could do the above ^^
> 
> OR fallback commit date or current date as last option.
> 
> I wonder why it started failing recently 🤔

No need to handle the date,I just removed it since it was never actually used in Acode anyway.
As for the recent failures, they started happening since the buildinfo plugin was added it just wasn’t reported until now. (this is a year old issue)

see:  https://verification.f-droid.org/packages/com.foxdebug.acode/

deadlyjack:
May be you can check the usage and implement the same methods in system plugin if possible or remove source for other platforms like windows, IOS as it's not being used.

UnschooledGamer:
Either Way implementing methods in system Plugin or not KEEP the LICENCE file & Credits along with them.

licaon-kter:
@UnschooledGamer we started making these viable recently

see: https://f-droid.org/2025/05/01/twif.html

and https://f-droid.org/2025/05/21/making-reproducible-builds-visible.html

PR #1405, opened 2025-07-01
chore(update IAP library to lateset version)

deadlyjack:


RohitKushvaha01 (commented):
1. Usage of querySkuDetailsAsync()
Noticed that billingClient.querySkuDetailsAsync() is being used in getProducts(). Since this method was removed in Play Billing Library 8.0.0, the recommended approach going forward is to use queryProductDetailsAsync() instead.

2. Calling enablePendingPurchases() without arguments
In getBillingClient(), the call to .enablePendingPurchases() is currently used without any parameters. Just to flag, this no-argument version was also removed in PBL 8.0.0

[Release notes](https://developer.android.com/google/play/billing/release-notes)

feridmoh:
Hi

PR #1426, opened 2025-07-15
All file access permission for terminal

RohitKushvaha01:
This PR fixes the 'Permission Denied' issue when accessing /sdcard and /storage/emulated/0 in terminal

Note:
The Play Store might reject this (don't blame me 🙃)

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -68,14 +68,6 @@
             <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
             <uses-permission android:name="android.permission.WRITE_MEDIA_STORAGE" />
             <uses-permission android:name="android.permission.VIBRATE" />
-
-            <!-- <uses-permission android:name="com.termux.permission.RUN_COMMAND" /> -->
-
-            <!-- Full file access (no prompts) -->
-            <!-- When granted, this permission allows Acode to access all files in internal and external storage without asking the user each time. -->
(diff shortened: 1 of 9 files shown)


UnschooledGamer:
Does the play store reject the whole app or specific version?

RohitKushvaha01:
> Does the play store reject the whole app or specific version?

I don't know, Play Store recommends not using it.

RohitKushvaha01:
Here's the documentation for context 
https://support.google.com/googleplay/android-developer/answer/10467955?hl=en

deadlyjack:
Make it optional and add a script which can toggle this feature when building. We'll now host one separate build on website with the permissions. We can try the new build on play store if ut gets we need to remove this permission and submit the app.

UnschooledGamer:
> Here's the documentation for context 
> https://support.google.com/googleplay/android-developer/answer/10467955?hl=en

Hmm, It says The App needs to specify the Reasons why it needs this permission in the declaration Form.

RohitKushvaha01:
Script uses:

```bash
node utils/storage_manager.mjs <y/n>

```

TECHNOBOT-OP:
You should also handle displaying of Internal Storage and external volumes in file browser in sdk 30 or above devices if Manage external Storage Permission is granted just as it is handled in lower versions

PR #1447, opened 2025-07-24
chore: update id-id.json

hyperz111:


UnschooledGamer:
Thx for updating. Ig Terminal still remains Terminal after translation 😉

PR #1459, opened 2025-07-27
Custom port support in Terminal manager

RohitKushvaha01:
Terminal is useless for plugins if we cannot connect to custom axs ports

### DIFF
--- a/src/components/terminal/terminal.js
+++ b/src/components/terminal/terminal.js
@@ -31,6 +31,7 @@ export default class TerminalComponent {
 			scrollOnUserInput: true,
 			rows: options.rows || 24,
 			cols: options.cols || 80,
+			port: options.port || 8767,
 			fontSize: terminalSettings.fontSize,
 			fontFamily: terminalSettings.fontFamily,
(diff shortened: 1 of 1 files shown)


UnschooledGamer:
I can't tell if Rohit, Is Describing a PR, or giving critics.

bajrangCoder (approved):

src/components/terminal/terminal.js
```diff
@@ -31,6 +31,7 @@ export default class TerminalComponent {
 			scrollOnUserInput: true,
 			rows: options.rows || 24,
 			cols: options.cols || 80,
+			port: options.port || 8767,
```
UnschooledGamer, line 34: I'm not sure but this might create injection vulnerabilities ig. Also it should be passed as number or whatever is expected type of it.

RohitKushvaha01: The purpose is to let plugin developers start terminal sessions that connect to their own AXS instances. So yes, it's technically an injection vulnerability but intended

PR #1517, opened 2025-08-17
Image optimization (reduced size by 24%)

sebastianjnuwu:
by @sebastianjnuwu


bajrangCoder:
We don't need to optimise every files. As the files which needs to be optimised is already optimised.

PR #1534, opened 2025-08-20
fix: big screen ui and global search worker issue

deadlyjack:
Changes:
- Fix big screen UI
  - Sidebar plugin app UI
  - Global search UI
- Fix worker not loading in global search
- Code cleanup

### DIFF
--- a/src/sidebarApps/extensions/style.scss
+++ b/src/sidebarApps/extensions/style.scss
@@ -64,28 +64,6 @@
         color: var(--secondary-text-color);
       }
     }
-
-    @media (min-width: 1024px) {
-      padding: 12px;
-
-      .title {
-        font-size: 1.25rem;
-
-        .icon-button {
-          padding: 0.5rem;
-          margin-left: 0.5rem;
-          border-radius: 8px;
-          min-width: 36px;
-          min-height: 36px;
-          font-size: 1.1em;
-        }
-      }
-
-      input[type="search"] {
-        padding: 0.6rem;
-        font-size: 0.95rem;
-      }
-    }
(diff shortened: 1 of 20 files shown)


src/sidebarApps/extensions/style.scss
```diff
@@ -131,7 +109,7 @@
     display: flex;
     justify-content: space-between;
     align-items: center;
-    min-height: 35px;
+    min-height: 30px;
```
bajrangCoder, line 112: 35px is fine for mobile screen, as with 30 the ui looks busy.

bajrangCoder: Also on wide screen , in extension tab there should be gap between icon and the text

bajrangCoder (changes requested):

src/sidebarApps/extensions/style.scss
deadlyjack: > 35px is fine for mobile screen, as with 30 the ui looks busy.

For now let it go, because 35px looks odd as parent has height of 30px.

> Also on wide screen , in extension tab there should be gap between icon and the text

yes, I'll add some padding

UnschooledGamer: Either way there's nothing to worry about 😄, May plugins like Better UI live long 🙂

bajrangCoder: > Either way there's nothing to worry about 😄, May plugins like Better UI live long 🙂

Still default should be good

Nehian97 (approved):

PR #1556, opened 2025-08-27
Apply font to the whole app

RohitKushvaha01:


bajrangCoder (approved):
👍

But idk if Robot mono font will look good on app, I think by default it should follow system one (for app ui)

bajrangCoder (approved):

RohitKushvaha01:
force pushed 🤦‍♂️

PR #1559, opened 2025-08-27
feat: exclude files and search exclude

alMukaafih:


alMukaafih:
Current implementation is too buggy when opening folders in files sidebar. It is extremely slow (in my own case up to 1 minute) for folders with large number of subfiles or subfolders.

jordanbrito2130-source:
Esto si dura

PR #1570, opened 2025-08-31
Terminal Service

### LINKED ISSUE #1569: Restoring Terminal session
### Check for existing issues

- [x] Completed

### Describe the feature

Basically, run the executor in a foreground service so that anything using the executor, like AXS, will survive the activity lifecycle and then reconnect to the AXS server to restore previous state

_No response_

RohitKushvaha01:
This change runs the Executor inside a foreground service, allowing it to survive the activity lifecycle. However, this does not prevent Android from killing Acode, nor does it preserve any UI state. 

Note: Any command started through the Executor or Terminal will continue running until it is manually stopped or terminated by the system.

Frontend changes are also required to reconnect AXS with the previously stored terminals, similar to how AcodeX handles reconnections.

Closes #1569 

# Todo
- [x] Foreground service
- [x] Ask for notification permission on api 35+
- [x] Exit button
- [x] WakeLock


RohitKushvaha01:
@bajrangCoder All that’s left is to restore the terminal tabs. axs will keep running even after the app is closed, as long as the service is running

PR #1590, opened 2025-09-16
Terminal initrc support

RohitKushvaha01:
### Summary

This PR adds support for an `/initrc` file, which is executed automatically whenever a new session is started. Users can customize this file to run commands at session startup, similar to a shell initialization script.

In addition, this script can now be used to change the default shell for the session.

### Changes

* Added support for `/initrc` script, executed at the start of each new session
* Fixed execution of `~/.bashrc`
* Corrected handling of the `$SHELL` environment variable

### DIFF
--- a/src/plugins/terminal/scripts/init-alpine.sh
+++ b/src/plugins/terminal/scripts/init-alpine.sh
@@ -2,7 +2,6 @@ set -e  # Exit immediately on Failure
 
 export PATH=/bin:/sbin:/usr/bin:/usr/sbin:/usr/share/bin:/usr/share/sbin:/usr/local/bin:/usr/local/sbin:/system/bin:/system/xbin:$PREFIX/local/bin
 export PS1="\[\e[38;5;46m\]\u\[\033[39m\]@localhost \[\033[39m\]\w \[\033[0m\]\\$ "
-export PIP_BREAK_SYSTEM_PACKAGES=1
 export HOME=/home
 export TERM=xterm-256color
 required_packages="bash"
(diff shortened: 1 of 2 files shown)


src/plugins/terminal/scripts/init-alpine.sh
```diff
@@ -29,21 +28,44 @@ if [[ ! -f /linkerconfig/ld.config.txt ]];then
...
     echo "$$" > $PREFIX/pid
     chmod +x $PREFIX/axs
-    $PREFIX/axs
+    
+    if [ -e "$PREFIX/alpine/initrc" ]; then
+        :
+    else
+        cat <<EOF > "$PREFIX/alpine/initrc"
+
+if [[ -f ~/.bashrc ]]; then
```
UnschooledGamer, line 64: Script Inconsistent 😶

src/plugins/terminal/scripts/init-alpine.sh
```diff
@@ -29,21 +28,44 @@ if [[ ! -f /linkerconfig/ld.config.txt ]];then
...
+    source ~/.bashrc
+fi
+
+if [[ -f /etc/bash/bashrc ]]; then
```
UnschooledGamer, line 68: Here too 🙃

src/plugins/terminal/scripts/init-alpine.sh
RohitKushvaha01: > Script Inconsistent 😶

image

UnschooledGamer (approved):

PR #1613, opened 2025-09-30
style(terminal): fix a mistyped `translate..()` CSS function

peasneovoyager2banana2:


bajrangCoder:
It make sense but previous one was also good

Issue #1452, opened 2025-07-26
Be able to add Acode-opened file shortcuts to the home screen


KeronCyst:
### Check for existing issues

- [x] Completed

### Describe the feature

1. Make a .TXT file to always be edited by Acode somewhere in your Droid's file system
2. Open Acode
3. Open the .TXT
4. Check the settings of the .TXT for an option to put a shortcut to the .TXT file directly on your home screen
5. Whenever you tap that .TXT file on your home screen, Acode launches and opens it

PR #1658, opened 2025-10-23
feat: add support for pinning file shortcuts to home screen

bajrangCoder:
Closes: #1452


RohitKushvaha01:
@bajrangCoder what about revoked uri permissions

RohitKushvaha01:
https://github.com/user-attachments/assets/e2630511-2d5e-45f1-b40f-393d78c9e7a9

And it doesn't work on hyperOs

bajrangCoder:
> @bajrangCoder what about revoked uri permissions

That will not open the app. What else can be done for it

PR #1682, opened 2025-11-10
fix: restore folds when formatting if available

bajrangCoder:


UnschooledGamer (approved):
The Implementation will work fine normally, But Incase where you had folded something like a method or a function, & the formatted code changes placement of them. This implementation may be problematic.

bajrangCoder:
> The Implementation will work fine normally, But Incase where you had folded something like a method or a function, & the formatted code changes placement of them. This implementation may be problematic.

In that cases there will some weird folding which will be obvious, what else we can do ? 
Btw for now we will keep this solution in Acode(ace)

UnschooledGamer:
> > The Implementation will work fine normally, But Incase where you had folded something like a method or a function, & the formatted code changes placement of them. This implementation may be problematic.
> 
> In that cases there will some weird folding which will be obvious, what else we can do ? Btw for now we will keep this solution in Acode(ace)

Could recalculate the folds based on the Updated Syntax Trees 🤔 💭

bajrangCoder:
> > > The Implementation will work fine normally, But Incase where you had folded something like a method or a function, & the formatted code changes placement of them. This implementation may be problematic.
> > 
> > 
> > In that cases there will some weird folding which will be obvious, what else we can do ? Btw for now we will keep this solution in Acode(ace)
> 
> Could recalculate the folds based on the Updated Syntax Trees 🤔 💭

But how will it going to detect where it need the recalculation

UnschooledGamer:
> > > > The Implementation will work fine normally, But Incase where you had folded something like a method or a function, & the formatted code changes placement of them. This implementation may be problematic.
> > > 
> > > 
> > > In that cases there will some weird folding which will be obvious, what else we can do ? Btw for now we will keep this solution in Acode(ace)
> > 
> > 
> > Could recalculate the folds based on the Updated Syntax Trees 🤔 💭
> 
> But how will it going to detect where it need the recalculation

By viewing the coords of old folds & the type/level of folds, and recreating that fold.

bajrangCoder:
> > > > > The Implementation will work fine normally, But Incase where you had folded something like a method or a function, & the formatted code changes placement of them. This implementation may be problematic.
> > > > 
> > > > 
> > > > In that cases there will some weird folding which will be obvious, what else we can do ? Btw for now we will keep this solution in Acode(ace)
> > > 
> > > 
> > > Could recalculate the folds based on the Updated Syntax Trees 🤔 💭
> > 
> > 
> > But how will it going to detect where it need the recalculation
> 
> By viewing the coords of old folds & the type/level of folds, and recreating that fold.

We can have some better solution in codemirror implementation

UnschooledGamer:
> > > But how will it going to detect where it need the recalculation
> > 
> > 
> > By viewing the coords of old folds & the type/level of folds, and recreating that fold.
> 
> We can have some better solution in codemirror implementation

@bajrangCoder, you're right. Implement the same functionality in it. Also, whenever you touch it once again. 
👍 for now

PR #1698, opened 2025-11-15
Update Portuguese translation

Gui64977:


Gui64977:
`preInstalled.js`: Changed `dark.popupActiveColor` (prev `rgb(66, 133, 244)`) to a better color (`rgb(219, 185, 0)`)

(`rgb(255, 215, 0)` might be better)

Before:


Now:

src/theme/preInstalled.js
```diff
@@ -15,7 +15,7 @@ dark.borderColor = "rgba(188, 188, 188, 0.15)";
 dark.popupIconColor = "rgb(245, 245, 245)";
 dark.popupBackgroundColor = "rgb(35, 39, 42)";
 dark.popupTextColor = "rgb(245, 245, 245)";
-dark.popupActiveColor = "rgb(66, 133, 244)";
+dark.popupActiveColor = "rgb(219, 185, 0)";
```
bajrangCoder: Using a different color will make it weird.
As this colour is not only used in completion popup

bajrangCoder (changes requested):

src/theme/preInstalled.js
Gui64977: Ok, thanks! I’ve reverted the change to `dark.popupActiveColor` as requested

Issue #1704, opened 2025-11-21
Create file or folder on FTP shows an error in File Browser 


icefront:
### Check for existing issues

- [x] Completed

### Describe the bug / provide steps to reproduce it

Menu -> Files
+ for creating a file or folder
Type the name and ok

ERROR: failed to connect to [ip] from [port] to ... after 60000 ms: isConnected failed: ECONREFUSED (Connection refused)

(Typo: the error message sentence begins with a lowercase letter)

Although the directory or file is created, it fails to refresh the view - it has to be closed and re-open.

The error appears immediately, not after 60000 ms.

### Environment

Samsung Galaxy S22 Ultra, Android 16

### If applicable, attach your Acode.log file to this issue.

<details><summary>Acode.log</summary><pre>

</pre></details>

PR #1706, opened 2025-11-23
fix(ftp): avoid doing nesting file/folder creation

bajrangCoder:
Closes: #1704


UnschooledGamer:
@bajrangCoder Have you checked the same for SFTP?

bajrangCoder:
> @bajrangCoder Have you checked the same for SFTP?

i wasn't able to fix that's why i removed in case of ftp but it works in case of sftp

Issue #1728, opened 2025-11-30
Backup more data?


blanhhy:
### Check for existing issues

- [x] Completed

### Describe the feature

Could you enhance the backup/restore feature to backup more data, such as theme, fonts, installed plugins?

I'm trying to migrate my Acode data to a new device, but app's backup/restore seems to only backup preferences.

I have used TitaniumBackup at first, but when I open Acode on new devices, it occurs too many errors and the app crashed quickly.

```log
java.io.FileNotFoundException: /data/user/0/com.foxdebug.acode/app_webview/webview_data.lock: 
open failed: EACCES (Permission denied)
```
(from `adb logcat` output)

bajrangCoder:
It already backup the settings, keybindings, plugins(not the local one), obviously theme , probably there will be any bug with custom font. But what else you want to backup? We can't back-up app state such as opened folders etc as they can't be restored because of saf uri workings etc.

That error is related to reading of some revoked permission files.

blanhhy:
@bajrangCoder 
I checked again and found that the backup file does contain all the data I wanted to back up. However, during the actual restoration process, it doesn't fully restore to a "ready-to-use" state.

Does the restoration of themes and plugin settings only take effect if the corresponding themes and plugins are already installed locally? If so, that feels a bit limited, because I'd have to manually check what I had installed before.

What I want exactly is either backing up the plugin local resources themselves, or having the restore feature automatically download (or at least prompt me what to download) any missing resources.

bajrangCoder:
It already installs the installed plugins after the restore

blanhhy:
@bajrangCoder no...it didn't install for me

PR #1744, opened 2025-12-14
fix: improve the backup and restore

bajrangCoder:
Fixes: #1728

### DIFF
--- a/utils/lang.js
+++ b/utils/lang.js
@@ -4,7 +4,9 @@ const yargs = require("yargs");
 const { hideBin } = require("yargs/helpers");
 const readline = require("node:readline");
 
-const args = yargs(hideBin(process.argv)).alias("a", "all").argv;
+const args = yargs(hideBin(process.argv))
+	.alias("a", "all")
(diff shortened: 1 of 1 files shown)


src/settings/backupRestore.js
```diff
@@ -115,93 +327,365 @@ function backupRestore() {
...
+		const fs = fsOperation(url);
+		let backupContent;
 
 		try {
-			backup = JSON.parse(backup);
+			backupContent = await fs.readFile("utf8");
 		} catch (error) {
-			alert(strings.error.toUpperCase(), strings["invalid backup file"]);
-			return;
+			throw new Error(`Could not read backup file: ${error.message}`);
```
UnschooledGamer, line 361: @bajrangCoder Translation string for this seems important

bajrangCoder: this will be just visible in logs file/console so there is no need for translation

UnschooledGamer: > this will be just visible in logs file/console so there is no need for translation

As you wish.

PR #1767, opened 2025-12-23
feat: Add smart path shortening to terminal prompt

bajrangCoder:
Screenshot_20251223-224649 Acode


RohitKushvaha01:
Are you sure it doesn't causes visual glitches in xterm.js? Because previous issue was caused by PS1

bajrangCoder:
> Are you sure it doesn't causes visual glitches in xterm.js? Because previous issue was caused by PS1

I haven't encountered any issues. You can check too

PR #1770, opened 2025-12-26
feat: handle preview webview permission by showing clear message about limitation

bajrangCoder:


src/plugins/browser/plugin.xml
RohitKushvaha01, line 1: This is bad and will raise serious privacy concerns (users and play store)

RohitKushvaha01 (changes requested):

src/plugins/browser/plugin.xml
sebastianjnuwu: There's no reason for a code-sharing app to have access to the microphone and camera.

bajrangCoder: Okay, I thought about that.

So instead of implementing these permission things , we can just handle it and show clear message to user instead of silently failing

UnschooledGamer: There are Reasons for needing them too, But not by default. They can be asked at runtime or whenever needed.

Take Examples from Zed, Vscode (Extension) using them for audio/video sharing in their collaboration stack.

bajrangCoder: > There are Reasons for needing them too, But not by default. They can be asked at runtime or whenever needed.
> 
> Take Examples from Zed, Vscode (Extension) using them for audio/video sharing in their collaboration stack.

Obviously it opens that native permission prompt when requesting for first time

PR #1783, opened 2025-12-31
feat(welcome): add welcome tab for first-time users

bajrangCoder:
- Add welcome tab and its command in command palette
- fix the issue of hideQuickTools parameter when creating custom editor tabs
- add docs url on help page
- keep social links on constant.js instead of here and there

### DIFF
--- a/src/lang/ar-ye.json
+++ b/src/lang/ar-ye.json
@@ -489,5 +489,9 @@
 	"issues found": "Issues found",
 	"error details": "Error details",
 	"active tools": "Active tools",
-	"available tools": "Available tools"
+	"available tools": "Available tools",
+	"recent": "Recent Files",
+	"command palette": "Open Command Palette",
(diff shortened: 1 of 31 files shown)


src/pages/welcome/welcome.js
```diff
@@ -0,0 +1,173 @@
...
+						shortcut="Ctrl+O"
+						onClick={() => acode.exec("open-folder")}
+					/>
+					<ActionRow
+						icon="historyrestore"
+						label={strings.recent}
+						onClick={() => acode.exec("recent")}
+					/>
+					<ActionRow
+						icon="tune"
```
UnschooledGamer, line 76: In the future, we may have to change this icon used for the command palette on the Welcome page. As the quick tools panel uses three horizontal dots.

bajrangCoder: yes but that doesn't look good on the page

PR #1798, opened 2026-01-05
Fix: Remove trimming of matched content in Executor.js

dikidjatar:
The output should not be trimmed. Using trim() breaks streamed output (line-by-line) because it removes leading spaces. This results in invalid output for many use cases, such as diffs and git status symbol.


RohitKushvaha01 (approved):
LGTM

PR #1802, opened 2026-01-06
feat: Offload file change detection to background thread

bajrangCoder:


src/plugins/system/android/com/foxdebug/system/System.java
```diff
@@ -616,6 +624,137 @@ private void encode(
...
+
+            Uri uri = Uri.parse(fileUri);
+            InputStream inputStream = null;
+            String fileContent;
+
+            try {
+                // Handle both file:// and content:// URIs
+                if ("file".equalsIgnoreCase(uri.getScheme())) {
+                    File file = new File(uri.getPath());
+                    if (!file.exists()) {
```
RohitKushvaha01: Also check file.canRead(). Sometimes invalid URIs get converted to /, which exists but is not what we expect.

Also check file.isFile() for the reasons stated above.

src/plugins/system/android/com/foxdebug/system/System.java
```diff
@@ -616,6 +624,137 @@ private void encode(
...
+            } finally {
+                if (inputStream != null) {
+                    try {
+                        inputStream.close();
+                    } catch (IOException ignored) {}
+                }
+            }
+
+            // check length first
+            if (fileContent.length() != currentText.length()) {
```
RohitKushvaha01, line 716: Move this check above the stream-reading code to avoid reading the file. Also, use file.length() instead of the file content length. (Only applicable for java.io.File)

src/plugins/system/android/com/foxdebug/system/System.java
```diff
@@ -616,6 +624,137 @@ private void encode(
...
+                // Read file content with specified encoding
+                Charset charset = Charset.forName(encoding);
+                byte[] bytes = new byte[inputStream.available()];
+                int totalRead = 0;
+                int bytesRead;
+                
+                // Read in chunks to handle large files
+                java.io.ByteArrayOutputStream buffer = new java.io.ByteArrayOutputStream();
+                byte[] chunk = new byte[8192];
+                while ((bytesRead = inputStream.read(chunk)) != -1) {
```
RohitKushvaha01: There are better ways to read from a stream/file, this is one of the most error prone method

For java.io.File you can just do `file.readText(encoding)`

bajrangCoder:
@greptileai review it

RohitKushvaha01 (approved):

PR #1831, opened 2026-01-18
feat: add eruda devtools for debugging

bajrangCoder:


src/main.js
```diff
@@ -219,6 +219,15 @@ async function onDeviceReady() {
 	acode.setLoadingMessage("Loading language...");
 	await lang.set(settings.value.lang);
 
+	if (settings.value.developerMode) {
+		try {
+			const devTools = (await import("lib/devTools")).default;
+			await devTools.init(false);
```
RohitKushvaha01, line 225: If someone enables devTools and doesn't use it, it loads eruda.js for no reason and wastes memory. It might be better to move the init call to the show() function

bajrangCoder: if it will not loaded from start then how it will capture logs

PR #1855, opened 2026-02-02
Arabic translation rework

makhlwf:
enhanced the arabic translation of the project

### DIFF
--- a/src/lang/ar-ye.json
+++ b/src/lang/ar-ye.json
@@ -19,7 +19,7 @@
 	"create folder error": "فشل إنشاء مجلد جديد",
 	"cut": "قص",
 	"delete": "حذف",
-	"dependencies": "المتطلبات",
+	"dependencies": "التبعيات",
 	"delay": "الوقت بالمللي ثانية",
 	"editor settings": "إعدادات المحرر",
 	"editor theme": "سمة المحرر",


src/lang/ar-ye.json
```diff
@@ -1,137 +1,137 @@
...
-	"lang": "(by Basel Al_hajeri & Hussain) العربية",
+	"lang": "العربية",
 	"about": "حول",
-	"active files": "الملفات مفتوحة",
+	"active files": "الملفات المفتوحة",
 	"add path": "إضافة مسار",
-	"alert": "تنبية",
-	"app theme": "لون التطبيق",
+	"alert": "تنبيه",
+	"app theme": "سمة التطبيق",
```
UnschooledGamer, line 7: The Updated Translation of `سمة التطبيق` translates to `Application feature` (on Google and Kagi translators), which is incorrect for this translation key.

src/lang/ar-ye.json
```diff
@@ -1,137 +1,137 @@
...
+	"commit message": "رسالة الـ Commit",
+	"console": "وحدة التحكم",
+	"conflict error": "خطأ! يرجى الانتظار قبل القيام بـ commit آخر.",
 	"copy": "نسخ",
-	"create folder error": "فشلت إنشاء مجلد جديد",
+	"create folder error": "فشل إنشاء مجلد جديد",
 	"cut": "قص",
 	"delete": "حذف",
-	"dependencies": "متطلبات",
+	"dependencies": "المتطلبات",
```
UnschooledGamer: Not sure, but `التبعيات` suits better here.

```suggestion
	"dependencies": "التبعيات",
```

src/lang/ar-ye.json
```diff
@@ -1,137 +1,137 @@
...
 	"delay": "الوقت بالمللي ثانية",
 	"editor settings": "إعدادات المحرر",
-	"editor theme": "لون المحرر",
```
UnschooledGamer, line 25: This translation seems better, Which translates as `Editor Color`, which is closer to the Translation Key of the App. Checked with Google Translate and Kagi. (Requires response from the Contributor)

makhlwf: Both are correct. I chose the word سمة because it aligns more with the word theme. However, if you like, you can apply your suggestion.

UnschooledGamer: > Both are correct. I chose the word سمة because it aligns more with the word theme. However, if you like, you can apply your suggestion.

That word translates (ref: kagi, google translate) as feature instead of color, theme or similar.

makhlwf: Yes, it could mean a feature if used in a different context. However, I checked before I translate. I've looked into other apps' translations. They use the same wording that I used. I even asked some AI if I used the wrong or the correct word or not, and they agreed that I used the correct one. However, as I told you, still, the word color works in this place. However, I see the word color is a more general word than theme, so this is why I choose the word theme. But still, using the word color still works.

UnschooledGamer: Got it. Seems like it was copying mistake as they sometimes get converted into different letters after copying them into the Clipboard. Sorry the Trouble Caused.

PR #1856, opened 2026-02-02
refactor: replace moment with dayjs

hyperz111:
`moment` is unmaintained, so i replace it with `dayjs`.


UnschooledGamer:
While Moment.js is unmaintained it's API is pretty much good considerably. The Moment.js is a Pretty Much Done on its base Feature set. Also, Moment.js explained this better in their docs here -> https://momentjs.com/docs/#/-project-status/

hyperz111:
I get the `moment` replacement from [this](https://e18e.dev/docs/replacements/moment.html)

bajrangCoder:
I don't think there is need for this change.

hyperz111:
> I don't think there is need for this change.

I get this build size difference.

Before:
```
$ du -sh www/build
5.9M    www/build
```

After:
```
$ du -sh www/build
5.6M    www/build
```

UnschooledGamer:
> I don't think there is need for this change.

@bajrangCoder, I think you should read https://momentjs.com/docs/#/-project-status/ before making a decision, also considering the package of moment.js (~4 mb) and dayjs being <1 mb

hyperz111:
@bajrangCoder, what about this?

UnschooledGamer:
@greptileai

UnschooledGamer:
@greptileai

UnschooledGamer (approved):
LGTM, Great.

PR #1888, opened 2026-02-20
Improved Splash Screen

RohitKushvaha01:


RohitKushvaha01:
@greptileai

bajrangCoder:
Update index.html , as it seems like it is using ace thing which got removed in cm one

PR #1912, opened 2026-02-27
Harden error handling and remove silent failures

bajrangCoder:
- Security/stability: no more swallowed runtime errors in critical paths (editor state, LSP, storage, ads, search navigation).
- Native plugin reliability: added logging for previously ignored exceptions.
- Better diagnostics: failures are now visible and actionable without crashing flows.


RohitKushvaha01 (approved):
LGTM, but someone will eventually complain about console being clogged with warns/errors

PR #1918, opened 2026-03-01
feat: rewarded ads thing so free user can also enjoy ad free

bajrangCoder:
- Removes the ad from opening any folder to sidebar
- Added a page where user can watch ads and enjoy uninterrupted ad free experience for certain time


RohitKushvaha01:
@bajrangCoder 
You don't need to copy ``EncryptedPrefrenceManager.java`` in ``system`` plugin you can just import it anywhere

```java
import com.foxdebug.acode.rk.auth.EncryptedPreferenceManager;
```

Issue #1745, opened 2025-12-15
`createFileStructure` does not pass file/folder intent to `fsOperation.extend` stat


Jobians:
https://github.com/Acode-Foundation/Acode/blob/d5b2ec94c710032d0e0a6c8efcd4fd7e8a98fcec/src/utils/helpers.js#L383

**Problem:**
`helpers.createFileStructure()` passes `isFile` only at the top-level call. Inside its loop, it calls:

```js
if (!(await fsOperation(newPath).exists())) { ... }
```

* `exists()` is called on `newPath`
* `createFile()` / `createDirectory()` is called on the parent path (`currentUri`)
* `stat()` inside `fsOperation.extend()` cannot know whether the path is intended to be a file or folder.

**Impact:**

* Files inside folders are treated as directories.
* Creating folders in the root may fail.
* Virtual paths cannot be correctly typed (`isFile` / `isDirectory`) before creation.

**Root cause:**
`isFile` is never passed to `fsOperation(newPath).exists()` or `stat()`.

**Proposed solution:**

* Expose the `isFile` intent to `fsOperation.extend()` / `stat()` calls inside the loop.
* Or provide a context/hook so FS extensions can detect the user’s intent.

**Notes:**

* Affects virtual file systems (paths not tied to a real filesystem).
* Current workarounds require storing `isFile` externally, which is complex.

Jobians:
Bump!

PR #1935, opened 2026-03-09
fix: resolve nested path creation from parent listings

bajrangCoder:
Fixes: #1745


bajrangCoder:
@greptileai review it again

PR #1938, opened 2026-03-10
chore: translate new strings for hu-hu.json

summoner001:
Translate new strings


UnschooledGamer:
@summoner001 Thanks for the translations. 🫡💚

PR #1990, opened 2026-03-28
fix(icons): clean up icon font usage and update the font

bajrangCoder:
- normalize icon references after merging li icons into icons.ttf
- restore inline alignment for plugin meta icons
- keep dragged tab icons at the correct size
- bust icons.ttf cache after glyph updates
- attached the icomoon project for easily adding new icons in future


bajrangCoder:
@greptileai

PR #2031, opened 2026-04-14
fix(file-browser): sync open folder tree after delete

### LINKED ISSUE #2028: Custom FS UI shows deleted file after create (lsDir not called after delete)
### Check for existing issues

- [x] Completed

In my custom filesystem (Acode.FS.extend), when I delete a file, Acode immediately removes it from the UI without calling lsDir().

However, after that:

If I create a new file, the UI shows both the new file AND the previously deleted file

The deleted file only disappears when I manually trigger or reopen lsDir()

This suggests Acode is updating the UI optimistically on delete instead of refreshing the directory from the filesystem.

---

🔁 Steps to reproduce

1. Open custom FS directory (bota://bots/{botId})

2. Load files via lsDir()

3. Delete a file using FS delete()

4. Observe: file disappears from UI (no lsDir() call happens)

5. Create a new file using FS createFile()

6. Observe: UI shows:

new file ✔

previously deleted file ❌ (ghost file)

7. Only fixed after manually triggering lsDir() again

---

❗ Expected behavior

After delete or create:

Acode should call lsDir() or fully refresh the directory

UI should always reflect actual filesystem state

---

❌ Actual behavior

delete() removes item from UI without re-reading FS

lsDir() is not called after delete

UI keeps stale directory state in memory

New file creation merges with stale list, causing ghost deleted entries

### Environment

nightly

https://github.com/user-attachments/assets/acd2e1f9-923e-4c9f-816b-f50fd3417a3f

_No response_

bajrangCoder:
Fixes: #2028

### DIFF
--- a/src/lib/openFolder.js
+++ b/src/lib/openFolder.js
@@ -586,7 +586,7 @@ function execOperation(type, action, url, $target, name) {
 		recents.removeFile(url);
 		if (helpers.isFile(type)) {
 			await fsOperation(url).delete();
-			$target.remove();
+			removeEntryFromOpenFolder(url);
 			const file = editorManager.getFile(url, "uri");
 			if (file) file.uri = null;
 			editorManager.onupdate("delete-file");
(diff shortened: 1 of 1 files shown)


bajrangCoder:
@greptileai

PR #2033, opened 2026-04-14
refactor(uk-ua.json): Translations

PavloPogonets:


UnschooledGamer:
@PavloPogonets Hey, is the Pull Request Ready, as I can see some strings untranslated?

PavloPogonets:
I did a final translation, but it seems to me that the line "Cargo crate" is not translated into Ukrainian.

UnschooledGamer:
> I did a final translation, but it seems to me that the line "Cargo crate" is not translated into Ukrainian.

Yes, that's right.

UnschooledGamer:
@PavloPogonets, is this mergeable, or may I convert this into a Draft Pull Request?

PavloPogonets:
I think it might be better to close it and I'll make a new translation because there are some errors here, there will probably be fewer problems.

UnschooledGamer:
> I think it might be better to close it and I'll make a new translation because there are some errors here, there will probably be fewer problems.

@PavloPogonets Alright, I'll close it.

PR #2053, opened 2026-04-20
Translation: Update hu-hu.json

summoner001:
Translate new strings.


summoner001:
What? ~20 string? I only modified just 2... Pls see my commit. Greptile wrong summary 😔

PR #2120, opened 2026-05-20
fix: plugin price display

deadlyjack:


deadlyjack:
@greptile_apps review

PR #2155, opened 2026-06-05
fix: terminal font dont always load

deadlyjack:


deadlyjack:
@greptile_apps review again

PR #2168, opened 2026-06-09
fix(editor): stabilize restored selection and loading state which was causing two edge cases

bajrangCoder:


bajrangCoder:
@greptileai

PR #2225, opened 2026-06-14
fix: improve cursor and quicktoolbar

deadlyjack:


deadlyjack:
@greptile_apps review again

PR #2234, opened 2026-06-14
feat(terminal): add terminal theme picker to Theme Settings page

deadlyjack:
- Move terminal theme selection from Terminal Settings to a new "Terminal" tab in Theme Settings with a live preview showing ANSI colors, sample prompt, and cursor blink animation
- Add preferredTerminalTheme to ThemeBuilder and assign matching terminal themes to all built-in app themes
- Auto-apply preferred terminal theme when switching app themes via themes.apply()
- Remove terminal theme field from Terminal Settings
- Fix: clear preview DOM before switching to terminal tab to prevent CodeMirror DOM leakage
- Fix: guard updateActiveTerminals() call against null editorManager in setTerminalTheme (matching safe pattern in theme/list.js)

### DIFF
--- a/src/pages/themeSetting/themeSetting.js
+++ b/src/pages/themeSetting/themeSetting.js
@@ -8,6 +8,7 @@ import { basicSetup, EditorView } from "codemirror";
 import Page from "components/page";
 import searchBar from "components/searchbar";
 import TabView from "components/tabView";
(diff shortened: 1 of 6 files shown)


deadlyjack:
@greptile_apps review again

PR #2235, opened 2026-06-14
fix(lsp): show dynamic install status label with loader feedback

deadlyjack:
- Display installation status as the primary label text instead of always showing "Installed", making it clear at a glance whether the server is installed, not installed, check-failed, or unknown.
- Wrap callback actions in a loader dialog for consistent visual feedback during async operations (install, update, uninstall).
- Add "lsp-status" translation key to all locale files.
- Fix missing "failed" case and non-existent "label-status" fallback in formatStatusLabel().


deadlyjack:
@greptile_apps review again

PR #2236, opened 2026-06-14
i18n(lang.js): Update Indonesian language name to Bahasa Indonesia

claycuy:


bajrangCoder:
what's the reason behind this change? (Note: I'm not native speaker of this language)

claycuy:
I changed it because 'Indonesian' usually refers to the citizens or nationality. In the native context, we call the language 'Bahasa Indonesia'. Using the native name (endonym) makes the language selection menu feel more natural and follows standard localization practices for local users.

PR #2240, opened 2026-06-15
feat(color-picker): add format dropdown, cancel button, and cancellat…

deadlyjack:
…ion handling

- Replace cycling format button with a dropdown popup (HSL/HEX/RGB)
- Add cancel button and reject promise on cancel/mask click
- Wire up cancellation handling in all 3 callers: commands.js, customTheme.js, and settingsPage.js
- Use CSS variable with fallback for format popup box-shadow
- Fix onhide callback to handle editor refocus on both OK and cancel

### DIFF
--- a/src/components/settingsPage.js
+++ b/src/components/settingsPage.js
@@ -574,11 +574,18 @@ async function resolveItemInteraction(item, $target) {
 		}
 
 		if (selectColor) {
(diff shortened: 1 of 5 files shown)


deadlyjack:
@greptile_apps review

PR #2244, opened 2026-06-15
refactor: replace url-parse to WHATWG URL API

hyperz111:
In `url-parse` readme:

> ...The URL interface is available in all supported Node.js release lines and basically all browsers. Consider using it for better security and accuracy.

So I think if We can do this for now :)


hyperz111:
WHATWG URL will throws on invalid input but url-parse not, maybe We should add exception handler?

RohitKushvaha01:
> WHATWG URL will throws on invalid input but url-parse not, maybe We should add exception handler?

Yes

hyperz111:
@greptile review again

hyperz111:
@greptile review again

hyperz111:
@RohitKushvaha01 what about the double-decode problem? Because the query values on original `url-parse` return value are also decoded.

RohitKushvaha01:
After taking a closer look at this PR, I think it's better not to merge it. While the changes seem reasonable, they could introduce hidden bugs besides the current implementation is working (sorry for the late review)

PR #2279, opened 2026-06-17
Update German translations in de-de.json

Mythryl-dev:
I updatet some of the roughter German translations to propper ones. I will continue at a later point to fix some more


bajrangCoder:
@greptile

Mythryl-dev:
It seems the npm check fails to locate the floating button string, as far as i can remmeber i changed it but did not delete it.

In case it is actually missing fill it with "Schwebende Schaltflächen"

Mythryl-dev:
Closed since more up to date version has been merged.

PR #2325, opened 2026-06-18
Better regex

Palloxin:
Better regex


Palloxin:
Fixed the problem the... summary pointed out.

Palloxin:
however, I cannot garantee they are 100% better or not-breakers without string contexts. For example ``/aa.*:/g`` might aim to match ONLY the very last ``:`` of the string.

If you have these regexes: 
1. ``/aa.*:/``
2. ``/aa.*?:/``

and this string:
``aaaaa:bbbb:cccc:``

The first one(1) will match ``aaaaa:bbbb:cccc:``

The second one(2) will match only
``aaaaa:``

The advantage of the lazyfication is that it prevents backtracking.
``/.+:/`` =>(``[^\n]+:``) eats everything until end string, then it backtracks(one step at time behind), checking if ``:`` is present (in front) at each step. It is a bigger waste of time the longer the string, and using ``.+?:`` is faster, unless you need the very last ``:`` of the string.

ANOTHER case where ``.+`` is better is when if you have a 100 meter long string, and you know(100%) the ``:`` you **want** is in the last 5-10 meters range, ``/.+:/`` is faster https://jsbench.me/ggmesp8ybc

RohitKushvaha01:
@greptileai

PR #2330, opened 2026-06-19
i18n: update Indonesian translation

claycuy:
## Features
- Readability/word enhancement
- Improvements for #2190 
- Update new string


claycuy:
@greptile review

claycuy:
@greptile review

claycuy:
@greptile

claycuy:
Are the options in cursor style and quicktools trigger mode not translatable?

RohitKushvaha01:
> Are the options in cursor style and quicktools trigger mode not translatable?

You can check the source code for that

claycuy:
If I want to make them translatable, am I allowed to add new i18n entries, or would that require a local APK build to verify?

RohitKushvaha01:
When adding a new i18n entry, it must be added to all language files, not just one. It's okay if the translations are missing for some languages, but the key itself must exist in every language file.

claycuy:
Okay, I'll give it a try. But I want to make sure, does that mean I need to modify the source code in `src/settings`?

PR #2379, opened 2026-06-24
fix: Use inline themed SVG for loader spinner

deadlyjack:
Import tail-spin.svg as raw SVG markup and render it through the loader DOM instead of using the circular border/background spinner. Update the SVG to inherit currentColor, expand its viewBox to avoid edge clipping, and style the dialog spinner with the app primary theme color.

Add ?raw SVG handling to both Rspack and Webpack so inline SVG imports do not conflict with normal asset/resource SVG usage. Also removes the old theme-side tail-spin recoloring path now that the spinner color is driven by CSS.

### DIFF
--- a/rspack.config.js
+++ b/rspack.config.js
@@ -89,9 +89,15 @@ module.exports = (env, options) => {
       ],
       type: 'javascript/auto',
     },
+    {
+      test: /\.svg$/,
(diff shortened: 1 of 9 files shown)


deadlyjack:
@greptile_apps review again

PR #2449, opened 2026-07-05
fix(editor): prevent styling loss flash on custom page tabs when splitting/closing panes

bajrangCoder:
- Implement DOM reconciliation in renderPaneLayout to keep pane element detachment to a minimum.
- Retrieve the document's main.css stylesheet and adopt it directly into the Shadow DOM.
- Catch adoption failures and fall back safely to a constructed CSSStyleSheet (caching only upon successful replaceSync compilation) or a link stylesheet.


bajrangCoder:
@greptile

Issue #2304, opened 2026-06-18
Syntax Highlight color flicker


Palloxin:
### Check for existing issues

- [x] Completed

### Describe the bug / provide steps to reproduce it

Html

https://github.com/user-attachments/assets/fe06cd43-7f0c-41d7-9924-e8a40fa9136d

The video has a 2nd phase where I reload the video at x0.2 speed to make it show better

just open an html and suddenly start typing. 

### Environment

idk

### If applicable, attach your Acode.log file to this issue.

<details><summary>Acode.log</summary><pre>

</pre></details>

bajrangCoder:
is rainbow bracket setting is enabled ? if so then it is fixed

Palloxin:
no the option is disabled. I made the videi with that option disabled since I dont really know what it does. More performance anyway.

bajrangCoder:
Pretty weird tho can't reproduce on latest build

Palloxin:
try with my settings.json
```json
{
    "animation": "yes",
    "appTheme": "glass dark",
    "autosave": 0,
    "fileBrowser": {
        "showHiddenFiles": false,
        "sortByName": true,
        "listFiles": true
    },
    "formatter": {},
    "prettier": {},
    "maxFileSize": 12,
    "serverPort": 8158,
    "previewPort": 8158,
    "showConsoleToggler": true,
    "previewMode": "inapp",
    "disableCache": false,
    "useCurrentFileForPreview": false,
    "host": "localhost",
    "search": {
        "caseSensitive": false,
        "regExp": false,
        "wholeWord": false
    },
    "lang": "en-GB",
    "uiZoom": 100,
    "fontSize": "11px",
    "cursorWidth": 2,
    "editorTheme": "tokyoNight",
    "textWrap": true,
    "softTab": false,
    "tabSize": 1,
    "retryRemoteFsAfterFail": false,
    "linenumbers": true,
    "formatOnSave": false,
    "fadeFoldWidgets": false,
    "autoCorrect": false,
    "openFileListPos": "header",
    "quickTools": 1,
    "quickToolsTriggerMode": "touch",
    "appFont": "",
    "editorFont": "Roboto Mono",
    "vibrateOnTap": false,
    "fullscreen": false,
    "floatingButton": false,
    "liveAutoCompletion": false,
    "localWordCompletion": false,
    "autoCloseTags": false,
    "autoRenameTags": false,
    "showPrintMargin": false,
    "printMargin": 80,
[...]

PR #2457, opened 2026-07-08
fix(editor): prevent syntax highlight color flicker on state recreation (#2304)

Elitex07:
Cached resolved language extensions (`file.__cmCachedLanguageExtension` and `file.__cmCachedLanguageSignature`) on the `EditorFile` instance. When `resolveLanguageExtension` runs for a file whose language signature matches the cache, it synchronously returns the cached extension immediately.


Elitex07:
@greptile

Elitex07:
@greptile

Elitex07:
@bajrangCoder :)

Palloxin:
does this fix only prevent the flicker or does it make the thing faster?

PR #2459, opened 2026-07-08
fix: implement editor tab history navigation (#1819)

### LINKED ISSUE #1819: Command for going to previously-used tab
### Check for existing issues

- [x] Completed

### Describe the feature

There are commands for "Open previous file tab" and "Open next file tab" which go back and forward in the list of tabs in the fixed order in which they're displayed. It would be useful to have a command that goes back to the tab opened before the current one. This is called "View: Open Previous Editor" in VS Code (`workbench.action.openPreviousEditorFromHistory"`). This way there would be an easy way to "jump back" after jumping to a different file (via "Go to definition" for example).

Running the command twice would go back two steps in history, etc. I guess there should be a command to go forward in history too. More details: https://www.geeksforgeeks.org/system-design/design-custom-browser-history-based-on-given-operations/

_No response_

Elitex07:
Implements editor tab visit history tracking and navigation commands (`prevFileHistory` / `nextFileHistory`), allowing users to traverse their open tabs in chronological order of visit rather than strictly left-to-right visual tab order.

Closes #1819

### DIFF
--- a/src/cm/commandRegistry.js
+++ b/src/cm/commandRegistry.js
@@ -364,6 +364,26 @@ function registerCoreCommands() {
 		},
 	});
 	addCommand({
+		name: "nextFileHistory",
+		description: "Open next file tab from history",
+		readOnly: true,
+		requiresView: false,
+		run() {
+			acode.exec("next-file-history");
+			return true;
+		},
+	});
+	addCommand({
+		name: "prevFileHistory",
+		description: "Open previous file tab from history",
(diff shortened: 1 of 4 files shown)


Elitex07:
@greptile

Elitex07:
@RohitKushvaha01

RohitKushvaha01 (approved):
Looks good on code, i will test it tomorrow because the GitHub action is failing

PR #2526, opened 2026-07-22
feat(file-tree): add performant virtualized file tree

bajrangCoder:


bajrangCoder:
@greptile

UnschooledGamer (approved):
LGTM, should use will change transform wherever it's not in the css style. As it would indicate the browser that this property(e.g: transform) may change later.

PR #2529, opened 2026-07-23
feat(files): move workspace indexing and search to native for saf/files

bajrangCoder:
## Summary

Moves file indexing and project-wide search for SAF and `file://` workspaces to the Android native layer.

This prevents large workspaces, such as Termux home, from creating a complete file tree in the WebView.

## Changes

- Added a native SQLite workspace index.
- Added paginated native filename queries.
- Added the new asynchronous `fileIndex` plugin API.
- Migrated Quick Open to native queries.
- Migrated project-wide search to native root-based search.
- Batched native search results to reduce Cordova bridge callbacks.
- Removed progress callbacks from individual file reads.
- Skips files and directories that return permission errors.
- Fixed project search after closing one folder and opening another.
- Fixed folder exclusion matching, including `node_modules`.
- Prevented recursive indexing of Termux `~/storage`.
- Kept FTP, SFTP, and custom storage providers on the JavaScript fallback.

## Breaking changes

`acode.require("fileList")` is deprecated.

It now contains files from non-native providers only. Plugins using it for SAF or `file://` workspaces must migrate to:

```js
const fileIndex = acode.require("fileIndex");
const result = await fileIndex.query({
  roots: [workspaceUrl],
  text: "filename",
  limit: 200,
});
```

Differences from `fileList`:

- `fileIndex` is asynchronous.
- Results are flat metadata records instead of `Tree` objects.
- Large results use cursor pagination.
- Native search can emit batched `search-results` events.

The native index database version is also updated. Existing generated indexes will be rebuilt automatically.

## Compatibility

- SAF and `file://`: native index and search.
- FTP and SFTP: existing JavaScript fallback.
- Low-level `sdcard.workspaceSearch()` keeps single-result events unless `batchResults: true` is provided.

### DIFF
--- a/src/components/fileTree/index.js
+++ b/src/components/fileTree/index.js
@@ -373,6 +373,8 @@ export default class FileTree {
 	 * @param {boolean} isDirectory
 	 */
 	appendEntry(name, url, isDirectory) {
+		if (this.entries.some((entry) => entry.url === url)) return;
+
 		const entry = { name, url, isDirectory, isFile: !isDirectory };
 
 		// Insert in sorted position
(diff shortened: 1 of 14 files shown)


bajrangCoder:
@greptile

PR #2533, opened 2026-07-24
feat(lsp): add document color chips with regex fallback

bajrangCoder:
Adds full LSP document color support (`textDocument/documentColor` + `textDocument/colorPresentation`) and keeps the existing regex color preview as a fallback for colors the language server misses.


bajrangCoder:
@greptile

bajrangCoder:
@greptile

PR #2537, opened 2026-07-26
feat(lsp): add pull diagnostics and native TypeScript 7

bajrangCoder:
- support pull diagnostics and server requests
- replace legacy TypeScript server with native TypeScript 7


bajrangCoder:
@greptile

PR #2566, opened 2026-08-02
fix: encrypt saved FTP/SFTP credentials

### LINKED ISSUE #2561: Remote server passwords (FTP/SFTP) are stored in cleartext in localStorage
### Check for existing issues

- [x] Completed

### What

When you add an FTP or SFTP server, Acode stores the connection in `localStorage.storageList` with
the username and password embedded directly in the URL, in cleartext.

The URL is built in `src/utils/Url.js` (`formate()`):

```js
// src/utils/Url.js:234
if (username && password) string += `${enc(username)}:${enc(password)}@`;
```

and the resulting record is saved to `localStorage` in `src/lib/remoteStorage.js` (`addFtp` around
line 58, `addSftp` around line 238):

```js
const res = { url, alias, name: alias, type: "ftp", home: null };
// url = ftp://user:password@host:port/...
```

So `localStorage.storageList` ends up holding entries like:

```
sftp://admin:SuperSecret123@10.0.0.5:22/
ftp://deploy:hunter2@example.com:21/
```

localStorage for the WebView is stored unencrypted on disk (the Chromium leveldb under
`app_webview/Default/Local Storage/`), so these passwords sit in plaintext at rest.

### Why it's worth fixing

I want to be clear about the threat model: on a normal, non-rooted device another app can't read
this, so this isn't remotely exploitable on its own. It matters as defense-in-depth:

- Anything with local access reads it in cleartext — a rooted/compromised device, `adb backup`, a
  lost/stolen device, or forensic extraction.
- Any future code-execution bug in the WebView gets the credentials for free. (This is exactly what
  happened with the Find-File XSS in GHSA-x24x-425w-326q — once JS ran in the WebView, one
  `localStorage.getItem("storageList")` returned every saved password. Fixing the storage removes
  that payoff even if another injection shows up later.)

### The part that makes this an easy call

Acode already does the right thing for its own account token — `shared_prefs/acode_auth_secure.xml`
uses AndroidX Security Crypto (`EncryptedSharedPreferences`, Tink). So the app already has an
encrypted-storage mechanism; the remote-server credentials just aren't using it.

### Suggested fix

- Don't keep the password in the URL string in `localStorage`. Store the non-secret parts
  (host, port, username, alias) where they are, and keep the password/passphrase/key in
  `EncryptedSharedPreferences` (or the Android Keystore), the same way the account token is handled.
- Or, minimally, encrypt the `storageList` blob at rest with a Keystore-backed key instead of writing
  it as plain JSON.

Happy to help test a change if that's useful.

### Version

Checked on v1.12.6 and current `main`.

### Environment

- Acode version: 1.12.6 (versionCode 1001)
- Android version: 17 (API 37)
- WebView: Chrome 145.0.7632.218
- Device: emulator (sdk_gphone64_arm64)

_No response_

</pre></details>

MYounas126:
Fixes #2561.

Saved FTP/SFTP servers are kept in `localStorage.storageList`, and the WebView writes localStorage to disk in plaintext — so the server passwords end up sitting unencrypted in `app_webview/.../Local Storage/leveldb`.

This PR moves that list into a native encrypted store (AES256-GCM via AndroidX Security-Crypto), which is the same mechanism the auth plugin already uses for the account token.

I started out planning to just pull the password out of the stored URL and keep it somewhere separate, but that doesn't work here: `Url.parse(url).url` keeps the password, and `helpers.getVirtualPath` matches active file URIs against that full URL to rewrite remote paths, so dropping the password breaks path display. So instead I left the in-memory data exactly as it was (the URL still has its credentials in memory, which is fine) and only changed where it's persisted — plaintext localStorage becomes an encrypted native store. That keeps the diff small and every existing reader behaves the same.

What's in here:

- A small `system.secureSet` / `secureGet` / `secureRemove` bridge (`SecureStore.java`, backed by `EncryptedSharedPreferences`).
- `src/lib/secureStorageList.js` — an in-memory cache loaded once at startup from the encrypted store, plus a one-time migration of any existing plaintext `localStorage.storageList`. The plaintext copy is only deleted after the encrypted write succeeds, so nobody loses their saved servers if something goes wrong mid-migration.
- The read/write sites in `Uri.js`, `helpers.js`, `externalFs.js` and `fileBrowser.js` now go through that module instead of touching `localStorage.storageList` directly.
- `onDeviceReady` awaits `secureStorageList.hydrate()` before anything reads the list.

Existing users' saved servers migrate automatically on the next launch — no re-entry.

I don't have the Cordova/Android build set up on my machine yet, so I haven't been able to run this on a device. Could you trigger a preview build, or tell me how you'd prefer it verified? The main things to check are: add an SFTP/FTP server, restart the app, and confirm it reconnects without asking for the password again — and that the credentials no longer appear in the localStorage leveldb.


MYounas126:
Thanks for the review! Pushed 57d8c16 addressing both points:

- Added the missing `<source-file>` declaration for `SecureStore.java` in plugin.xml so Cordova compiles it (System.java references it).
- Switched `SecureStore` from `apply()` to `commit()` and made `secure-set` report failure, so the migration only removes the plaintext storageList after the encrypted write is durably on disk — no data loss if the process is killed mid-migration.

UnschooledGamer:
@greptile review again.

src/lib/secureStorageList.js
```diff
@@ -0,0 +1,97 @@
+/**
+ * Secure persistence for the remote-storage list (saved FTP/SFTP servers).
+ *
+ * Historically the list lived in `localStorage.storageList`, which the WebView
+ * writes to disk in cleartext — so every saved server password sat unencrypted
+ * in `app_webview/.../Local Storage/leveldb`. See issue #2561.
```
bajrangCoder: Is it necessary to mention these

UnschooledGamer:
> I don't have the Cordova/Android build set up on my machine yet, so I haven't been able to run this on a device. Could you trigger a preview build, or tell me how you'd prefer it verified? The main things to check are: add an SFTP/FTP server, restart the app, and confirm it reconnects without asking for the password again — and that the credentials no longer appear in the localStorage leveldb.

No problem. Changes look safe for a possible preview trigger.

src/lib/secureStorageList.js
MYounas126: Good point — trimmed it down to just the essentials in 07ae2584. Dropped the historical/localStorage-path detail and kept only what's needed to understand the module (encrypted store, sync cache, hydrate at boot).

src/plugins/system/android/com/foxdebug/system/SecureStore.java
```diff
@@ -0,0 +1,76 @@
...
+        try {
+            String masterKeyAlias = MasterKeys.getOrCreate(MasterKeys.AES256_GCM_SPEC);
+            prefs = EncryptedSharedPreferences.create(
+                PREF_NAME,
+                masterKeyAlias,
+                context,
+                EncryptedSharedPreferences.PrefKeyEncryptionScheme.AES256_SIV,
+                EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM
+            );
+        } catch (GeneralSecurityException | IOException e) {
```
bajrangCoder, line 46: **The fallback format can cause data loss after a transient encryption failure.**
This writes the literal storageList key to ordinary preferences and reports success, causing the JS migration to delete localStorage. If encryption works after the next restart, EncryptedSharedPreferences encrypts lookup keys and cannot see that literal plaintext key, so the server list appears empty. The fallback needs its own migration path/status, or encryption failure should propagate so the legacy copy remains. Android confirms that this implementation encrypts both keys and values. [Android documentation](https://developer.android.com/reference/androidx/security/crypto/EncryptedSharedPreferences)

MYounas126: Great catch — you're right, and it's worse than just data loss: the fallback also silently re-introduced cleartext. Fixed in e57354e3 by removing the plaintext fallback entirely. Now if the encrypted store can't be opened, `set`/`get` return false/null instead, so `secure-set` reports failure and the migration keeps the legacy localStorage copy to retry next launch — nothing is ever written unencrypted, and there's no invisible-key scenario.

bajrangCoder:
@greptile

UnschooledGamer:
A side note: this can break any plugins that use storageList from `localStorage`!

MYounas126:
Good point, thanks for flagging it. A couple of thoughts:

`storageList` isn't part of the public `acode` plugin API (it's not in `acode.define(...)`), so any plugin reading it is going through `localStorage` directly — an internal detail rather than a supported surface.

And there's a security angle to that same access: right now any installed plugin can read `localStorage.storageList` and pull every saved SFTP/FTP password in cleartext. Moving it into the encrypted store removes that, which is really part of the point here — untrusted plugin code shouldn't have free read access to the user's server credentials.

That said, if there are plugins that legitimately need the saved-server list, I'm happy to add a proper accessor to the `acode` API that returns just the non-secret fields (alias/host/type, no password) so those keep working while credentials stay protected. Do you know of any plugins relying on it, or an API shape you'd prefer?

bajrangCoder:
> Good point, thanks for flagging it. A couple of thoughts:
> 
> `storageList` isn't part of the public `acode` plugin API (it's not in `acode.define(...)`), so any plugin reading it is going through `localStorage` directly — an internal detail rather than a supported surface.
> 
> And there's a security angle to that same access: right now any installed plugin can read `localStorage.storageList` and pull every saved SFTP/FTP password in cleartext. Moving it into the encrypted store removes that, which is really part of the point here — untrusted plugin code shouldn't have free read access to the user's server credentials.
> 
> That said, if there are plugins that legitimately need the saved-server list, I'm happy to add a proper accessor to the `acode` API that returns just the non-secret fields (alias/host/type, no password) so those keep working while credentials stay protected. Do you know of any plugins relying on it, or an API shape you'd prefer?

Yeah, there is no need for a plugin to read.

UnschooledGamer:
@MYounas126 Well, yeah. That's the usage for these types things comes from edgey cases. Security is first, While I cannot link any plugins code related it as I'm not in a state to do so currently. It might be fine to move forward; As it was just a side note.

UnschooledGamer:
The legitimate plugin I found other than deprecated ones due to our CM 6 Migration:

https://github.com/overskul/acode-better-filebrowser/blob/a83aaf8b4d50c12b6130141faef04b58fc6e5f81/src/BetterFileBroswer.js#L1-L98

Also, With the proposed implementation the plugins can still access them just like Acode can as long as the Java bridge provides a decoded, decrypted to the Javascript side.

UnschooledGamer:
> Yeah, there is no need for a plugin to read.

@bajrangcoder Well, no. As a Plugin needs know which folders the App has access to, which helps for URI restructuring, `addedFolder` doesn't provide that completely as far as I know.

MYounas126:
Thanks, went and looked at better-filebrowser — fair point, it's a real case. It reads and writes storageList but only touches uuid/name/uri (no creds, it's just registering a local folder), so my change would break it since I'm pulling storageList out of localStorage entirely.

And you're right about the bridge as well. secureGet is reachable from any JS in the origin, so a plugin can still read the encrypted list, passwords included, the same way Acode does. So realistically all this PR buys is getting the cleartext copy off disk, which was the original #2561 concern (backups, forensics, other apps reading the webview data dir). It doesn't keep the list away from plugins, and I don't think that's really doable in a single-origin webview anyway.

So maybe a cleaner approach: keep the non-secret list in localStorage so those plugins keep working (uuid/name/uri/type), and only move the passwords into the encrypted store, keyed per entry and put back at connect time. That still fixes the at-rest issue, doesn't break the plugins, and actually keeps passwords out of the plugin-visible list. It's more work than what's here now — I avoided it at first because the password-in-url matching in getVirtualPath gets in the way, but that's fixable by stripping creds off both sides before comparing.

I'm happy to redo it that way, or keep this one as the at-rest fix and do the split as a follow-up. Whatever you two think is best.

MYounas126:
@UnschooledGamer @bajrangCoder  I am waiting for your reply. Once I got your suggestion I will modify my strategy accordingly. Please
Regards.

UnschooledGamer:
CC: @bajrangCoder , as I'm busy currently & unable to decide that.

bajrangCoder:
> So maybe a cleaner approach: keep the non-secret list in localStorage so those plugins keep working (uuid/name/uri/type), and only move the passwords into the encrypted store, keyed per entry and put back at connect time. That still fixes the at-rest issue, doesn't break the plugins, and actually keeps passwords out of the plugin-visible list. It's more work than what's here now — I avoided it at first because the password-in-url matching in getVirtualPath gets in the way, but that's fixable by stripping creds off both sides before comparing.

Looks fine

MYounas126:
> > So maybe a cleaner approach: keep the non-secret list in localStorage so those plugins keep working (uuid/name/uri/type), and only move the passwords into the encrypted store, keyed per entry and put back at connect time. That still fixes the at-rest issue, doesn't break the plugins, and actually keeps passwords out of the plugin-visible list. It's more work than what's here now — I avoided it at first because the password-in-url matching in getVirtualPath gets in the way, but that's fixable by stripping creds off both sides before comparing.
> 
> Looks fine

Pushed the reworked version in f7e8fe81 (also merged main in, the branch had drifted).

Rebuilt it the way we discussed, so this is quite different from what was here before:

**The saved-server list stays in localStorage.** `storageList` keeps `uuid`, `name`, `url`, `type`, `home` exactly as before, so better-filebrowser and anything else reading it keeps working. I reverted the changes that had moved the whole list out.

**Only the secrets move.** Passwords and SFTP key passphrases go into the encrypted store (`system.secureSet`/`secureGet`, backed by EncryptedSharedPreferences), and the saved URL is written without the `:password@` part. So localStorage no longer holds credentials, but everything else about the list is untouched.

**Secrets go back in at connect time.** `Ftp.fromUrl` and `Sftp.fromUrl` already call `Url.decodeUrl(url)`, so that turned out to be a clean hook if the URL has no password, they pull it from the store. `remoteStorage.edit()` does the same so the edit dialog still prefills.

**getVirtualPath.** As expected this needed the fix you mentioned: it now strips the password off both sides before comparing. Without that, URLs saved before the migration (which still contain `user:pass@`) stop matching once new ones don't.

**Migration.** On startup it scans `storageList` for URLs with inline credentials, moves them into the encrypted store, then rewrites the entry without the password. The localStorage copy is only rewritten after the encrypted write is confirmed on disk if that fails it keeps the old copy and retries next launch, so nobody loses their servers mid-migration.

New file is `src/lib/secureCredentials.js`; `secureStorageList.js` is deleted.

---

One thing I'd like your opinion on: I key the stored secrets by connection identity (`protocol//user@host:port`) rather than by the entry's `uuid`. Reason is that `fromUrl` only ever receives a URL, it has no reference to the storageList entry, so keying by uuid would mean threading extra context through the fs layer.

The trade-off is that two saved entries pointing at the same user@host:port share one secret. That's almost always the same credential anyway, but it does mean editing the password on one entry changes it for the other. If you'd rather have strict per-entry isolation I can rework it to pass the uuid down, just wanted to flag it rather than have you find it in review.

Also worth being upfront: I still don't have the Cordova/Android build running locally, so this is lint-clean but I haven't run it on a device. The two paths worth exercising are adding an SFTP/FTP server then restarting to confirm it reconnects without asking for the password again, and upgrading from a current build to confirm existing servers survive the migration. Happy to get another preview build if that's easiest.

bajrangCoder:
@greptile

src/lib/secureCredentials.js
```diff
@@ -0,0 +1,171 @@
...
+		/^([a-z0-9+.-]+:\/\/)([^@/]*?):([^@/]*)@/i,
+		(_, scheme, user) => `${scheme}${user}@`,
+	);
+}
+
+/** Promisified bridge helpers — resolve to null instead of throwing. */
+function secureGet(key) {
+	return new Promise((resolve) => {
+		try {
+			window.system.secureGet(key, resolve, () => resolve(null));
```
bajrangCoder, line 52: window.system.secureGet and secureSet return Promises and do not accept callbacks, so the callbacks passed here are ignored. This outer Promise never resolves during normal operation, leaving hydrate() and therefore onDeviceReady() : stuck indefinitely.

src/lib/remoteStorage.js
```diff
@@ -232,10 +237,14 @@ export default {
...
 			await helpers.showInterstitialIfReady();
+
+			// Keep secrets in the encrypted store instead of the saved URL (#2561).
+			await secureCredentials.set(url, { password, passPhrase });
+
 			return {
 				alias,
 				name: alias,
-				url,
+				url: secureCredentials.stripPassword(url),
```
bajrangCoder, line 247: stripPassword() only removes the userinfo password; the URL constructed above still contains passPhrase in its query string. updateStorage later serializes this URL into plaintext localStorage, so newly saved key-based SFTP passphrases remain exposed. Build the persisted URL without passPhrase or strip that query parameter as well.

src/lib/secureCredentials.js
```diff
@@ -0,0 +1,171 @@
...
+		const key = keyFor(url);
+		if (!key) continue;
+
+		const password = decodeURIComponent(
+			/^[a-z0-9+.-]+:\/\/[^@/]*?:([^@/]*)@/i.exec(url)?.[1] || "",
+		);
+		if (!password) continue;
+
+		pending[key] = { ...(pending[key] || {}), password };
+		entry.url = stripPassword(url);
```
bajrangCoder, line 116: This condition only admits URLs containing user:password@, and the migration only extracts that password. Legacy key-authenticated SFTP entries commonly have no inline password but do have passPhrase in the query, so they are skipped and remain plaintext.

src/lib/secureCredentials.js
```diff
@@ -0,0 +1,171 @@
...
+	await secureSet(SECURE_KEY, JSON.stringify(next));
+	cache = next;
+}
+
+/**
+ * Drop stored secrets for a connection (used when a server is removed).
+ * @param {string} url
+ */
+async function remove(url) {
+	await set(url, {});
```
bajrangCoder, line 168: This removal API has no callers. fileBrowser.removeStorage deletes the list entry and key file without invoking it, while editing a connection identity writes the new key without deleting the old one. As a result, credentials survive deletion and identity changes indefinitely.

bajrangCoder (changes requested):

bajrangCoder:
@MYounas126 I have a different solution which is profiles, you can check #2694

MYounas126:
> @MYounas126 I have a different solution which is profiles, you can check #2694

Had a proper look at #2694, and your approach is better than mine. A few things it does that mine doesn't:

**The URL carries nothing.** `sftp://profile-<uuid>/path` leaves no hostname, username or port in localStorage at all. Mine only strips the password, so `user@host:port` is still sitting there in plaintext.

**Secrets never reach JS.** This is the one that actually decides it. In my version `system.secureGet` is callable from any JS in the origin, so a plugin can still read the passwords back out I flagged that myself earlier in this thread and didn't have a good answer for it. Your profile reads return metadata only and the secrets never cross the bridge, which closes it properly.

**Key files stop being written to disk.** I hadn't touched that at all. The current code copies the private key into DATA_STORAGE in cleartext and mine leaves that as-is; yours reads it natively and keeps it inside the encrypted profile. That's arguably a bigger exposure than the password in localStorage was.

Plus Keystore-managed AES/GCM with a per-profile IV and the profile ID as AAD is stronger than what I was doing with EncryptedSharedPreferences, and you have tests for it.

So I'd rather see #2694 go in than mine. Happy to close this one.

One thing before I do, in case it's useful: #2694 only covers SFTP as far as I can tell  no FTP files in the diff, and `addFtp` in remoteStorage.js is untouched, so it still does `Url.formate({ protocol: "ftp:", username, password, ... })` and the FTP password still ends up in plaintext localStorage. So #2561 wouldn't be fully closed by merging #2694 on its own.

If you want, I can do a follow-up that applies the same profile pattern to FTP once #2694 lands, so both sides end up consistent rather than FTP keeping the old URL format. Or if you'd rather handle it yourself as part of #2694, that's fine too  just flagging it so it doesn't get missed.

Either way, thanks for the reviews on this one. The points you raised (the callback/promise mismatch that would have hung startup, the passphrase still being in the query string, the migration skipping key-auth entries, and the unused remove path) were all real, and I learned a fair bit from them.

bajrangCoder:
Yes, you're right that #2694 doesn't touch any ftp stuffs and that is intentional cuz that pr is supposed to improve the SFTP/ssh things and putting all together will make things hard for reviewers.

Also your striping up things from url method was error-prone and so we'll go for similar approach for ftp too(but after the merge of #2694). As secret stuffs should not come to js thread at all otherwise plugin can do malicious stuffs.

PR #2567, opened 2026-08-02
Improve recent dialog path readability

bajrangCoder:
## Summary

Improves the recent files and folders dialog so entries remain identifiable on mobile

## Changes

- Display the file or folder name as the primary text
- Display its location as secondary text
- Show complete paths when space is available
- Truncate overflowing paths from the left, preserving the useful ending
- Decode SAF document paths into readable locations
- Preserve absolute Termux paths and their leading `/`
- Use configured storage aliases for concise primary labels
- Increase the dialog width responsively
- Improve spacing and remove-button touch targets
- Expose complete paths through accessibility labels and hover titles
- Extend select-dialog items with subtitles and custom styling hooks
- Add regression coverage for primary storage and Termux SAF URIs

### DIFF
--- a/src/dialogs/select.js
+++ b/src/dialogs/select.js
@@ -11,13 +11,17 @@ import restoreTheme from "lib/restoreTheme";
  * @property {string} [default]
  * @property {function():void} [onCancel]
  * @property {function():void} [onHide]
+ * @property {string} [className]
  */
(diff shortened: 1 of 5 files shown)


bajrangCoder:
@greptile

PR #2587, opened 2026-08-03
fix(terminal): close tab on process end and speed up long-press select

bajrangCoder:
Close the terminal tab when the PTY exits or the socket drops, not only when the AXS exit JSON arrives, so users no longer get a dead zombie tab after typing exit. Also lower the default touch selection hold from 600ms to 400ms so long-press select feels more responsive.


bajrangCoder:
@greptile

PR #2621, opened 2026-08-04
fix(editor): prevent false language mode recommendations

bajrangCoder:
- recommend only language-mode plugins available in the registry
- suppress built-in, extensionless, and arbitrary file false positives
- remove the direct plugin-request issue action
- add regression tests for unknown extensions


bajrangCoder:
@greptile

PR #2623, opened 2026-08-04
fix(lsp): send workspace folders during initialization

bajrangCoder:
Pass the resolved workspace root through initialize.workspaceFolders so language servers such as ty can discover project configuration and Python virtual environments.


bajrangCoder:
@greptile

PR #2632, opened 2026-08-04
Feat/lsp improvements

gat0sy:
# Summary

Improves Acode's built-in CM6 LSP client with fixes for URI handling, workspace edits, client capabilities, session management, and custom server configuration. All changes are client-side; no external tooling or server scripts are included.

### Position & text edit handling

- Fixed "applyTextEdits" discarding formatter/code-action results when an LSP server returns "end.line == doc.lines" (the standard end-of-document convention).
- Added "safeLspPositionToOffset" to safely clamp document positions without throwing.

### URI resolution

- Fixed forward URI translation for Acode's internal storage provider ("foxdebug.acode" / "foxdebug.acodefree") so internal-storage projects are correctly mapped to "file://" URIs.
- Added reverse URI translation (LSP "file://" → Acode URI) for internal storage, allowing features such as Go to Definition and tooltip file links to work correctly.
- Fixed SFTP URI translation in both directions so remote workspaces resolve consistently.

### Client capabilities

- Declared missing client capabilities ("workspace.applyEdit", "workspace.workspaceFolders", "textDocument.codeAction.resolveSupport") so servers such as jdtls and Dart can advertise and execute workspace edits and code actions.

### Workspace edits

- Generalized workspace/applyEdit handling in transport.ts so server-initiated workspace edits can be applied across multiple files through AcodeWorkspace, with per-file success/failure reporting.

### LSP actions

- Added a dedicated LSP actions menu exposing existing and supported LSP features such as Go to Definition, Show References, Rename Symbol, and Code Actions.
- Automatically executes code actions when only a single action is available.

### Session lifecycle

- Added a 45-second idle grace period before disposing inactive LSP clients, avoiding unnecessary restarts while switching files.

### Custom server configuration

- Replaced the multi-step "add_custom_server" wizard with pre-filled prompts.
- Added an Edit action for custom servers so transport, launcher settings, and language IDs can be modified without recreating the server.

### Tooltip file links

- Fixed an Android "FileUriExposedException" when opening "file://" links from hover or signature-help tooltips. Links now open directly inside Acode at the correct location if accessible by the latter.

## Testing

- "biome check --write" ✓
- Tested with jdtls, Dart Analysis Server, phpactor, pylsp, and jedi-language-server over WebSocket transport.
- Verified on shared storage, internal storage, and SFTP workspaces.


gat0sy:
@deadlyjack Separately from this PR, I wanted to share an idea I have been experimenting with for SFTP + LSP support.

The way I got SFTP projects working reliably was by running the LSP on the same machine as the files, then forwarding the WebSocket connection back to Acode through SSH. This avoids a lot of URI/path translation issues because the language server sees the real filesystem it is indexing.

The "acode-ls" script I shared earlier was not actually Termux-specific. It only had a Termux shebang because I used Termux as my testing environment when the built-in terminal was unavailable.

> Failed to create terminal:
> Executor.setProotDebug is not a function

 The same concept works with Acode's own WebSocket LSP setup.

The idea would be to make this more seamless:

- Detect when the opened workspace is an SFTP workspace.
- Use the existing SSH credentials to open a port-forwarded session.
- Check if the remote machine already has the WebSocket bridge/LSP runtime available.
- If not, optionally offer to install it.
- Start the remote LSP bridge automatically and connect Acode to it.

Then when the SFTP workspace is closed, the session can be stopped.

The goal is not to replace local LSPs, but to make heavy servers usable remotely. For example, jdtls is often too heavy for a phone, but running it on the same machine as the project worked perfectly for me: indexing, completion, code actions, imports, generation of getters/setters, rename, etc.

The general idea is: the LSP should run where the files are. Local projects use local servers, SFTP projects use remote servers through SSH forwarding.

I'm not sure if this fits Acode's roadmap, but I thought I would share the approach since SFTP support already exists and this could make large projects much more practical on mobile devices. Happy to discuss it or drop it if it is out of scope.

gat0sy:
A merge has been done to stay in check with upstream, the last commit fixes LspOffsetToPosition into a SafeLspOffsetToPosition that uses the clamping method, it may be worth taking a look at it some time in the furture when possible.

gat0sy:
@bajrangCoder

This PR should bring LSP references, Go to Definition, and Symbol Rename to all supported workspace types: built-in Alpine, external storage (SAF), and SFTP. The only remaining gaps should be a few edge cases that I still need to test.

One idea I had for SFTP: since it already uses SSH, we could reuse the existing connection to automatically start the remote WebSocket bridge and forward it back to Acode instead of requiring users to launch it manually. I added a "forwardPort" helper to "sftp.js" as preparation for this, but didn't wire it up yet since I wanted to keep this PR focused.

The main benefit is that LSPs would run on the same machine as the project, which avoids URI translation issues and lets index-heavy servers (like JDTLS) work much better over SFTP. The tunnel could simply follow the workspace lifecycle and shut down automatically when the session ends.

The next things I'll look at (in separate PRs) are:

- ~Hover tooltips on split panes (currently missing on secondary editor views)~ 
- Hover not triggering with mouse cursor moves over elements.

That should cover the core LSP improvements I had planned. Let me know if you'd like anything changed or split differently.

I will share some screen recordings of the features at work later.

bajrangCoder:
> One idea I had for SFTP: since it already uses SSH, we could reuse the existing connection to automatically start the remote WebSocket bridge and forward it back to Acode instead of requiring users to launch it manually. I added a "forwardPort" helper to "sftp.js" as preparation for this, but didn't wire it up yet since I wanted to keep this PR focused.

> The main benefit is that LSPs would run on the same machine as the project, which avoids URI translation issues and lets index-heavy servers (like JDTLS) work much better over SFTP. The tunnel could simply follow the workspace lifecycle and shut down automatically when the session ends.

Yes, We have also planned about this but for now we'll first improve the core the sure in future, SFTP(ssh) LSP support will be added(but optional)

bajrangCoder:
@greptile

src/cm/lsp/codeActions.ts
```diff
@@ -156,9 +151,10 @@ async function applyChangesToFile(
 	workspace: AcodeWorkspace,
 	uri: string,
 	changes: LspChange[],
-	mapping: { mapPosition: (uri: string, pos: Position) => number },
+	mapping: { mapPosition: (uri: string, pos: lsp.Position) => number },
```
bajrangCoder, line 154: This references `lsp.Position`, but no `lsp` namespace is imported in this file. Btw previously it was correct why are you even changing it

bajrangCoder:
why this pr feels like slop cuz i can see there are several edits which are just useless and irrelevant

Please don't do this, if anything got missed while reviewing then it will be hard to track.

I'm closing it, make new pr with proper title and try not to change whole things in codebase(unless needed). Make separate PRs for fixes, new things instead of putting everything at once(Also make sure to test/check codes before making pr)

PR #2700, opened 2026-08-08
feat(lsp): fixing LspToPosition via clamping and moving it to helper

gat0sy:
Fix lspPositionToOffset when LSP servers return EOF positions one line past the document via clamping

Extract lspPositionToOffset() and applyTextEdits() into a shared helper so clientManager and transport use the same bounds-safe logic of the new lspPositionToOffset

Send workspaceFolders when rootUri is available

Declare missing applyEdit, workspaceFolders, and code action resolve capabilities

Re-send workspace/didChangeConfiguration after initialization

Normalize SFTP URIs before sending them to LSP servers ( Assuming the LSP runs on the same machine as the sftp directory )

These fixes improve compatibility with servers that rely on workspace folders, applyEdit, resolvable code actions, post-init configuration, or full-document edits. They may need some polishing.

Note: CodeActions feature times out if more than one lsp are running at the same time when using the built-In LSPs

### DIFF
--- a/src/cm/lsp/clientManager.ts
+++ b/src/cm/lsp/clientManager.ts
@@ -9,7 +9,7 @@ import {
   serverCompletion,
   serverDiagnostics,
 } from "@codemirror/lsp-client";
-import { EditorState, Extension, Facet, MapMode } from "@codemirror/state";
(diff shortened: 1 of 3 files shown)


bajrangCoder:
@greptile

PR #2715, opened 2026-08-11
refactor(plugin install): speed up plugin installation with native ZIP extraction

codewithchai0605:
## Summary

This PR improves plugin installation performance, especially for plugins containing many small files such as file/folder icon providers.

The previous installer extracted archive entries in JavaScript with a concurrency limit of 2. This caused significant overhead from repeated JS-to-native filesystem calls.

This change moves archive extraction to native Android code and writes files directly from the ZIP archive.

## Changes

- Extract plugin ZIP archives natively using `ZipFile`.
- Stream archive entries directly to a staging directory.
- Activate updates through a directory swap, preserving the existing plugin if extraction fails.
- Restore the previous plugin installation if an earlier update was interrupted during the swap.
- Prevent concurrent installations of the same plugin.
- Validate plugin IDs and archive paths to prevent path traversal.
- Add archive size and entry-count limits.
- Optimize extraction for many small files:
  - shared 64 KiB buffer
  - direct output streams
  - no per-entry canonical filesystem lookup

### DIFF
--- a/src/lib/installPlugin.js
+++ b/src/lib/installPlugin.js
@@ -197,7 +197,11 @@ export default async function installPlugin(
 				plugin,
 			);
 			loaderDialog?.setMessage("Extracting plugin files...");
-			await extractPluginArchive(archiveUrl, pluginDir, JSON.stringify(pluginJson));
+			await extractPluginArchive(
+				archiveUrl,
+				pluginDir,
+				JSON.stringify(pluginJson),
+			);
 			extractionComplete = true;
 
 			if (isDependency) {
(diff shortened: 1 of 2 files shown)


codewithchai0605:
@greptile-apps review

codewithchai0605:
@greptile-apps review

RohitKushvaha01 (changes requested):
Why did you change Tee.java? It's for security related stuff this doesn't make any sense put that code in System.java

codewithchai0605:
@greptile-apps review

PR #2728, opened 2026-08-12
fix(terminal): stop Ctrl+Shift+V pasting twice

simyo:
Trying to paste with a physical keyboard using Ctrl+Shift+V in the built-in terminal pastes the clipboard content twice. Copy and font-zoom (Ctrl+/-) have the same issue.

**Cause:** `terminal.js`'s `setupCopyPasteHandlers` uses xterm.js's `attachCustomKeyEventHandler`, which fires on both `keydown` and `keyup`. The handler never checked `event.type`, so `pasteFromClipboard()` / `copySelection()` / font-zoom ran once per phase — twice per keypress.

**Fix:** only invoke the action on `keydown`; still `preventDefault()`/swallow the event on both phases so nothing leaks through to default handling.

**Verification:** *Works On My Machine™*

### DIFF
--- a/src/components/terminal/terminal.js
+++ b/src/components/terminal/terminal.js
@@ -439,17 +439,22 @@ export default class TerminalComponent {
 	setupCopyPasteHandlers() {
 		// Add keyboard event listener to terminal element
 		this.terminal.attachCustomKeyEventHandler((event) => {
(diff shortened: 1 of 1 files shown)


simyo:
Thanks for your support, Mr. Bot.

Issue #2738, opened 2026-08-14
always show cancel btn On SFTP folder loading


Victozee26:
### Check for existing issues

- [x] Completed

### Describe the feature

<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/98b0e3d3-378a-4efb-a773-25b1f50968bb" />

when host is offline is usually loads for 5 - 7secs before showing the "Cancel" button which is not comfortable UX

i suggest always showing the Cancel btn or

on Acode launch 
if previous interacted folder type = SFTP
change File Browser cwd = /

<img width="708" height="509" alt="Image" src="https://github.com/user-attachments/assets/92f30934-1cc7-41c4-ad30-257309543471" />

THANKS FOR ACODE 🙏

deadlyjack:
fix in pr #2743

PR #2743, opened 2026-08-14
fix: issue #2738

deadlyjack:
show cancel button instantly when loading files for remote files.


deadlyjack:
@greptileai review again

PR #2752, opened 2026-08-15
fix(android): harden edge-to-edge release builds

deadlyjack:
## Summary

- Remove the Android edge-to-edge opt-out while preserving current inset-safe layouts.
- Route system-bar styling through modern APIs on Android 15+ and keep legacy color compatibility below API 35.
- Add inset-safe CrashActivity handling for system bars and display cutouts.
- Enable release-only R8 optimization and resource shrinking with conservative Cordova, plugin, SSH, and crypto keeps.
- Add one focused unit test covering the opt-out, release shrink configuration, keep coverage, and the narrow SSHTools JMX exception.

## R8 note

Maverick Synergy has an optional desktop thread-dump path that references three java.lang.management classes unavailable on Android. The rules name exactly those three classes; there is no wildcard dontwarn, ignorewarnings, dontoptimize, or dontobfuscate rule. Acode does not enable maverick.threadDump.

## Validation completed

- npm run test: 33 files, 239 tests passed
- npm run typecheck
- biome ci .
- paid production AAB and APK, signed and minified
- free production AAB and APK with AdMob, signed and minified
- non-empty mapping files embedded in both AABs
- AAB/APK ZIP integrity and signatures verified
- package IDs and launcher/notification fallback resources verified
- DEX output: paid 8,380,120 bytes; free 12,553,664 bytes; local baseline 19,572,936 bytes

## Remaining acceptance gates

- [ ] Android 14, 15, and 16 manual matrix: gesture and three-button navigation, portrait/landscape, light/dark, keyboard, and cutout profiles
- [ ] Manual feature matrix for editor/files, BrowserActivity, fullscreen WebView, crash UI, SAF, terminal notification, network plugins, IAP, and free AdMob flows
- [ ] Free AAB Play internal-track upload and clean pre-launch report
- [ ] Staged rollout monitoring for startup, plugin loading, crashes, and ANRs

This stays in draft until the remaining device and Play gates pass. The connected Android 16 phone contains Play-signed installs, while local artifacts use the upload key; replacing them would require deleting app data. The local API 36 emulator image is incomplete, so device testing was not forced.

### DIFF
--- a/build-extras.gradle
+++ b/build-extras.gradle
@@ -2,6 +2,15 @@ android {
     packagingOptions {
         pickFirst 'META-INF/versions/9/OSGI-INF/MANIFEST.MF'
     }
+
+    buildTypes {
+        release {
+            minifyEnabled true
+            shrinkResources true
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'),
+                rootProject.file('../../proguard-rules.pro')
+        }
+    }
 }
 
 configurations {
--- a/proguard-rules.pro
+++ b/proguard-rules.pro
@@ -1,0 +1,22 @@
+# Keep Cordova's reflection-loaded bridge and Acode's native plugin code stable.
+-keep class org.apache.cordova.** { *; }
+-keep class com.foxdebug.** { *; }
+-keep class com.silkimen.** { *; }
+-keep class com.verso.** { *; }
+-keep class admob.plus.** { *; }
+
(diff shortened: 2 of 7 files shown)


RohitKushvaha01:
I think this PR is a bit too ambitious and has a fairly large scope, which could make it more difficult to review and may increase the risk of introducing hidden bugs. It would be better to break it down into smaller, focused PRs so they’re easier to review and test.

PR #2761, opened 2026-08-17
added remove method for selectionMenu, add method now return an item

vikramgaming:


bajrangCoder:
What's the point of changing add method stuffs

vikramgaming:
@bajrangCoder sometimes when I make plugins, we can add a selectionMenu but we cannot remove it.

so what if the add method return the menuItem and that menuItem can be used to remove the selectionMenu from items?

Sorry if my English is bad, I'm using a translate app

PR #2781, opened 2026-08-20
fix: fall back to internal storage when external storage is unavailable

ulite-Amr:
## What

Fixes the app failing to start on devices where the external storage directory cannot be created. On such devices users were stuck on the splash screen with a misleading **"Acode failed to start. Update Android System WebView or Chrome."** message — the WebView was fine, the problem was storage.

## Reported by

- https://t.me/foxdebug_acode/1/157932
- https://t.me/foxdebug_acode/150650/164093
- https://t.me/foxdebug_acode/1/71463

## Root cause

1. `src/main.js` unconditionally picked the external data directory:
   `window.DATA_STORAGE = externalDataDirectory || dataDirectory;`
2. The Cordova `file` plugin only registers `files-external` if the external root directory can be created (`newRoot.mkdirs()` fails on affected devices), so the chosen URL points at a non-existent filesystem.
3. The subsequent plugins-directory creation was **not** wrapped in try/catch:
   `if (!(await fsOperation(PLUGIN_DIR).exists())) await fsOperation(DATA_STORAGE).createDirectory("plugins");`
4. Its unhandled rejection aborted `onDeviceReady()` before `loadApp()`, leaving `editorManager` null and the splash screen forever — with the error handler blaming WebView/Chrome.

Logcat evidence on an affected device:

```
FileUtils: Unable to create root dir for filesystem "files-external", skipping
FileUtils: Unable to create root dir for filesystem "cache-external", skipping
Uncaught (in promise) #<FileError> at build/main.js (63221)
Uncaught (in promise) TypeError: Cannot read properties of null (reading 'hasUnsavedFiles')
```

## Changes

- **`src/main.js`**
  - Added `resolveStorageDir()` — probes the preferred directory with `fs.stat()` and falls back to internal storage when unavailable
  - Wrapped plugins-directory creation in try/catch; on failure it reassigns `DATA_STORAGE`/`CACHE_STORAGE`/`PLUGIN_DIR`/`KEYBINDING_FILE` to internal storage and retries
  - Storage permission requests are now guarded: only requested on SDK < 33, and only when `hasPermission` returns false (`ensurePermission()` helper), preventing background-activity-launch blocks
- **`src/lib/acode.js`** — `editorManager?.hasUnsavedFiles?.() ?? 0` in `exitAppMessage` (the error above was a downstream symptom of the aborted startup)
- **`www/index.html`** — startup error/rejection handlers now detect storage errors (`FileError`/`NotFoundError`/`SecurityError`/DOMException codes 1-12) and show an honest **"Acode failed to start: storage is unavailable"** message instead of blaming WebView/Chrome

## Tested

The same patch was verified on an affected device: app now falls back to internal storage and starts normally, and storage errors produce an accurate message. CI (biome, typos, vitest) runs on this PR.

Fixes the class of startup failures reported as "Update Android System WebView or Chrome" that actually originate from unavailable external storage.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -141,12 +141,25 @@
         );
       }
 
+      function isStorageError(reason) {
+        if (!reason) return false;
+        return (
+          reason.name === "FileError" ||
+          reason.name === "NotFoundError" ||
+          reason.name === "SecurityError" ||
+          (typeof reason.code === "number" && reason.code >= 1 && reason.code <= 12)
+        );
+      }
+
       function handleStartupError(event) {
         if (!isStartupLoading()) return;
         var message = event && event.message ? event.message : "Startup error";
         setStartupMessage(
-          "Acode failed to start. Update Android System WebView or Chrome. " +
-          message
+          isStorageError(event && event.error)
+            ? "Acode failed to start: storage is unavailable. Try restarting your device, or clearing the app data. " +
+                message
+            : "Acode failed to start. Update Android System WebView or Chrome. " +
+                message
         );
       }
 
@@ -156,8 +169,11 @@
         var message =
           reason && reason.message ? reason.message : "Startup promise failed";
         setStartupMessage(
-          "Acode failed to start. Update Android System WebView or Chrome. " +
-          message
+          isStorageError(reason)
+            ? "Acode failed to start: storage is unavailable. Try restarting your device, or clearing the app data. " +
+                message
+            : "Acode failed to start. Update Android System WebView or Chrome. " +
+                message
         );
       }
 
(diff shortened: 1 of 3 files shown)


RohitKushvaha01 (approved):

www/index.html
```diff
@@ -141,12 +141,25 @@
         );
       }
 
+      function isStorageError(reason) {
+        if (!reason) return false;
+        return (
+          reason.name === "FileError" ||
+          reason.name === "NotFoundError" ||
+          reason.name === "SecurityError" ||
+          (typeof reason.code === "number" && reason.code >= 1 && reason.code <= 12)
```
bajrangCoder: Treating every numeric code from 1 through 12 as a storage error is weird, it can be different like dom or cordova plugin errors

bajrangCoder (approved):

PR #2805, opened 2026-08-25
fix(android): avoid splash surface crash on realme Android 13

bajrangCoder:


hooks/post-process.js
```diff
@@ -210,6 +211,79 @@ function enableStaticContext() {
...
+    }
+
+    @Override
+    protected void loadConfig() {
+        super.loadConfig();
+
+        // Some realme Android 13 builds pass a null SurfaceControl while handing the
+        // splash view to an app-provided exit animation. Avoid only that handoff.
+        if (hasBrokenRealmeSplashTransfer()) {
+            preferences.set("FadeSplashScreen", false);
```
RohitKushvaha01, line 264: Everything else looks fine but are you sure the ``preferences`` variable is defined? And guaranteed to be not null

RohitKushvaha01 (approved):

hooks/post-process.js
bajrangCoder: Yes. `preferences` is a protected field from `CordovaActivity`. This override calls `super.loadConfig()` first, which assigns it from `ConfigXmlParser#getPreferences()`. That parser eagerly initializes its preferences field with `new CordovaPreferences()`, so it is non-null even if parsing returns early. Cordova also dereferences it immediately after `loadConfig()` without a null check.

PR #2817, opened 2026-08-27
fix(plugin-install): Move plugin zip extraction to native (Java) code, remove hardcoded 2-file batch limit

codewithchai0605:
## Problem

`installPlugin.js` extracted plugin zip archives on the JS thread in
hardcoded batches of 2 files, yielding to the UI thread with
`await new Promise(r => setTimeout(r, 0))` between batches. This was
presumably to stop the WebView from freezing during install, but it
made installing plugins with many files unnecessarily slow, and the
batch size had no principled basis.

## Fix

Move the actual unzip + disk I/O to native Java code, running on
Cordova's background thread pool instead of the JS thread. Since
native code never touches the UI thread, there's nothing to protect
by batching — the whole archive is extracted in a single native call.

### New plugin: `src/plugins/pluginInstaller`

- `PluginInstaller.java` — opens the (already-downloaded, on-disk) zip
  with `ZipFile`, walks entries, sanitizes each path (mirrors the
  existing `sanitizeZipPath` / `isUnsafeAbsolutePath` logic from
  `installPlugin.js` so zip-slip protection isn't lost), writes files
  with plain `java.io.File` I/O, and computes SHA-256 checksums in the
  same format `installState.js` already uses, so existing install
  state remains valid.
- `PluginInstaller.js` (www bridge) — exposes `extractZip()` /
  `cancelExtract()` over `cordova.exec`.
- `plugin.xml` / `package.json` — standard Cordova plugin scaffolding,
  following the same pattern as the existing `pluginContext` plugin.

### `installPlugin.js` changes

- The old batching loop, and the `createFileRecursive` /
  `sanitizeZipPath` / `isUnsafeAbsolutePath` helpers it used, are
  replaced by a single call to `PluginInstaller.extractZip()`.
- The downloaded archive is staged as a temp file under
  `CACHE_STORAGE` and native code streams from that path, rather than
  base64-encoding the whole thing through the JS bridge (which would
  otherwise inflate the payload ~33% and hold several full copies of
  it in memory across JS/bridge/Java).
- Extraction progress is reported back and shown in the loader
  dialog.
- The loader's cancel button is now actually wired up: cancelling
  aborts the in-flight download (where the underlying transport
  supports it) or the in-flight extraction, instead of just closing
  the dialog while the install continues in the background.
- A single bad/oversized archive entry is skipped and logged instead
  of aborting the entire install.

### Hardening included along the way

- Zip-slip protection is preserved, plus a cheap, purely lexical
  containment check as a second line of defense (no per-file
  filesystem/symlink-resolution syscalls).
- Zip-bomb guards: a cap on any single entry's decompressed size and
  on the total decompressed size of the archive, since this code path
  extracts arbitrary user-supplied/downloaded archives.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -113,6 +113,7 @@
         "com.foxdebug.acode.rk.exec.proot": "file:src/plugins/proot",
         "com.foxdebug.acode.rk.exec.terminal": "file:src/plugins/terminal",
         "com.foxdebug.acode.rk.plugin.plugincontext": "file:src/plugins/pluginContext",
+        "com.foxdebug.acode.rk.plugin.plugininstaller": "file:src/plugins/pluginInstaller",
         "cordova-android": "^15.1.0",
         "cordova-clipboard": "^1.3.0",
         "cordova-plugin-acode-webview": "file:src/plugins/webview",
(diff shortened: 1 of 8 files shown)


codewithchai0605:
@greptile-apps review

codewithchai0605:
@greptile-apps review

codewithchai0605:
@greptile-apps review

RohitKushvaha01:
Thanks but this pr has too many issues and the performace impact is minimal so have no plans of changing it

PR #2829, opened 2026-08-29
perf(file-browser): Migrate `cachedDir` to `Map` to prevent V8 hidden class degradation

AuDevTist1C:
# #2831


bajrangCoder:
remove the invented-looking benchmark table and GC/V8 claims.(the whole pr title + body is just slop there is nothing meaningfull)

I'm closing this pr open new with real pr title , body instead of these slop otherwise I'll block

PR #2525, opened 2026-07-22
feat: add WebView Plugin API

RohitKushvaha01:
Add cordova-plugin-acode-webview allowing plugins to create and manage isolated WebView instances with fullscreen, window, panel, and hidden modes.

Features:
- Create independent WebView instances via acode.webview.create()
- Load URLs, HTML content, and execute JavaScript
- Bidirectional messaging between plugin and WebView
- Lifecycle management: show, hide, reload, destroy
- Lifecycle events: pageFinished, titleChanged, closed

### Pre/Post Merge To-do:

- [x] Plugin types updation.
- [x] Plugin Docs updation.

Closes #2281

PR #2836, opened 2026-08-30
fix(webview): retain instance when destroy fails

azamkassim:
Summary

- Mark a WebView instance destroyed only after native destruction succeeds.
- Keep the JavaScript instance usable when native destruction fails.
- Add focused tests for failed, retried, and successful destruction.

Why

The WebView API currently sets "_destroyed" before awaiting the native bridge. If native destruction fails, the native WebView may still exist, but its JavaScript wrapper permanently rejects further operations—including another destruction attempt.

This change moves the state transition until after successful native destruction.

Testing

- "npm test" — 49 test files and 331 tests passed
- "npm run typecheck"
- Biome check passed
- "git diff --check" passed

Related work

Follow-up to the WebView Plugin API introduced in Acode-Foundation/Acode#2525.

### DIFF
--- a/src/lib/webview.js
+++ b/src/lib/webview.js
@@ -126,8 +126,8 @@ class WebView {
 
 	async destroy() {
 		this._checkDestroyed();
-		this._destroyed = true;
 		await nativeBridge.destroy(this.id);
+		this._destroyed = true;
 		instances.delete(this.id);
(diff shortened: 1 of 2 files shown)


bajrangCoder:
@greptile

PR #2841, opened 2026-08-31
feat: httpStream

### LINKED ISSUE #2827: Native HTTP: stream response bodies so plugins can consume SSE / fetch() ReadableStream
### Check for existing issues

- [x] Completed

### Describe the feature

`cordova.plugin.http.sendRequest` currently buffers the entire response before the success callback. Plugins that talk to LLM APIs (OpenAI, Anthropic, OpenRouter, Gemini, xAI, …) therefore cannot do real HTTP SSE.

Please add a **streaming body** path to the native HTTP plugin. Native code should not parse SSE. It should deliver bytes/text as they arrive so JS can wrap the result as a WHATWG `Response` with a `ReadableStream` body.

_No response_

RohitKushvaha01:
Closes #2827

```js

const res = await system.httpStream("https://httpbin.org/stream/5");
const reader = res.body.getReader();
const dec = new TextDecoder();
while (true) {
  const { done, value } = await reader.read(); // value: Uint8Array
  if (done) break;
  const chunk = dec.decode(value, { stream: true });
  console.log(chunk);
}

```


RohitKushvaha01:
@greptile

RohitKushvaha01:
@greptile

RohitKushvaha01:
@greptile

RohitKushvaha01:
@bajrangCoder check now

src/plugins/system/android/com/foxdebug/system/StreamHttp.java
```diff
@@ -0,0 +1,400 @@
...
+      Log.w(TAG, "Failed to release stream callback " + requestId, e);
+    }
+  }
+
+  @Override
+  public void run() {
+    try {
+      if (cancelled) return;
+      URL url = new URL(this.url);
+      connection = (HttpURLConnection) url.openConnection();
```
bajrangCoder: **[bug]** `cancel()` can fail to stop the in-flight request. `connection` is a plain field (not volatile, not assigned under `creditLock`), and `run()` only re-checks `cancelled` after the request has already been configured and the body written. Two failures follow: (1) a cancel that wins before the assignment sees `connection == null` and never calls `disconnect()`, so `connect()`/`getOutputStream()`/`getResponseCode()` keep running until the connect timeout; (2) even when `cancelled` is already true, `run()` still opens the connection and writes a POST body. That contradicts the documented guarantee that aborting the signal or cancelling the reader cancels the underlying native HTTP request.

**Suggestion:** Publish `connection` under the same lock `cancel()` uses (or make it volatile and assign-then-recheck). After `openConnection()`, if `cancelled` is set, disconnect immediately and return. Check `cancelled` again before writing the body, and disconnect from `cancel()` while holding that lock so a connection assigned concurrently cannot be missed.

src/plugins/system/android/com/foxdebug/system/StreamHttp.java
```diff
@@ -0,0 +1,400 @@
...
+    // Set-Cookie, which cannot be comma joined) survive the JSON bridge.
+    JSONArray headerPairs = new JSONArray();
+    if (headerFields != null) {
+      for (Map.Entry<String, List<String>> entry : headerFields.entrySet()) {
+        String name = entry.getKey();
+        List<String> values = entry.getValue();
+        if (name == null || values == null || values.isEmpty()) {
+          continue;
+        }
+        String lower = name.toLowerCase();
```
bajrangCoder: **[bug]** Response header names are lowercased with `name.toLowerCase()`, which uses the device default locale. On Turkish/Azeri locales, ASCII `I` becomes U+0131 (`ı`), so names such as `X-Request-ID` or `If-None-Match` become `x-request-ıd` / `ıf-none-match`. `Headers.append` then stores a different header than the caller will look up with `headers.get("x-request-id")`.

**Suggestion:** Use `name.toLowerCase(Locale.ROOT)` (and import `java.util.Locale`).

src/plugins/system/android/com/foxdebug/system/StreamHttp.java
```diff
@@ -0,0 +1,400 @@
...
+      event.put("chunk", Base64.encodeToString(chunk, Base64.NO_WRAP));
+    }
+    sendEvent(event, true);
+  }
+
+  // True when the chunk has no byte below 0x20, i.e. nothing the JSON encoder
+  // is forced to escape (as a U+XXXX sequence) into a larger payload.
+  private static boolean jsonSafe(byte[] chunk) {
+    for (byte b : chunk) {
+      if ((b & 0xFF) < 0x20) return false;
```
bajrangCoder: **[suggestion]** `jsonSafe` rejects any byte `< 0x20`, including `\n`/`\r`/`\t`. SSE frames (the motivating use case in #2827) always contain newlines, so every LLM/SSE chunk takes the base64 path — ~33% larger bridge payloads plus `atob` on the JS side. That is strictly worse than the latin1 path: Android `JSONObject` already escapes those three bytes as `\n`/`\r`/`\t` (2 chars), which still round-trips through `latin1ToBytes`.

**Suggestion:** Treat 0x09/0x0A/0x0D as json-safe (or only fall back to base64 for bytes that JSON would encode as `\u00XX`). Keep 0x00–0x08 / 0x0B / 0x0C / 0x0E–0x1F on the base64 path.

src/plugins/system/android/com/foxdebug/system/StreamHttp.java
```diff
@@ -0,0 +1,400 @@
...
+    }
+  }
+
+  @Override
+  public void run() {
+    try {
+      if (cancelled) return;
+      URL url = new URL(this.url);
+      connection = (HttpURLConnection) url.openConnection();
+      connection.setRequestMethod(method);
```
bajrangCoder: **[suggestion]** The method string is passed straight to `HttpURLConnection.setRequestMethod`. That API is case-sensitive and only allows `GET`, `POST`, `HEAD`, `OPTIONS`, `PUT`, `DELETE`, and `TRACE`. Fetch-style `method: "post"` and `PATCH` (used elsewhere in this repo via `src/lib/ajax.js` and `cordova-plugin-advanced-http`) throw `ProtocolException`, which is reported as a transport failure.

**Suggestion:** Normalize with `method.toUpperCase(Locale.ROOT)` in `httpStreamStart` or `run()`. For `PATCH` (and any other non-standard verb), use the usual Android workaround (reflection on the methods list, or `X-HTTP-Method-Override`) rather than letting `setRequestMethod` fail.

src/plugins/system/www/plugin.js
```diff
@@ -281,5 +281,231 @@ module.exports = {
...
+              var headers = headersFromPairs(event.headers || []);
+              var response;
+              if (cannotHaveBody) {
+                response = new Response(null, {
+                  status: status,
+                  statusText: event.statusText || '',
+                  headers: headers
+                });
+              } else {
+                response = new Response(stream, {
```
bajrangCoder: **[suggestion]** `new Response(...)` is not wrapped in try/catch. The constructor throws `RangeError` if `status` is outside 200–599 (HttpURLConnection returns `-1` for a malformed status line) and can also throw if `statusText` is not a valid reason-phrase. That exception escapes the Cordova success callback, so the returned promise hangs, the native stream keeps running, and nothing calls `http-stream-cancel`.

**Suggestion:** Wrap Response construction in try/catch, `fail(err)` (which rejects or errors the body), and `sendCancel()` so a bad status cannot leak a native request.

src/plugins/system/www/plugin.js
```diff
@@ -281,5 +281,231 @@ module.exports = {
...
+   *   is a normal response (not a rejected promise); only transport failures
+   *   reject. Cancelling the returned stream's reader (or aborting
+   *   `options.signal`) cancels the underlying native HTTP request.
+   */
+  httpStream: function (url, options) {
+    options = options || {};
+    var signal = options.signal || null;
+
+    var nativeOptions = {};
+    for (var key in options) {
```
bajrangCoder: **[suggestion]** `nativeOptions` is a shallow copy of `options` minus `signal`. A fetch-like `headers: new Headers([["Authorization", "Bearer x"]])` JSON-serializes as `{}`, so the native request is sent with no headers. The public types say `Record<string, string>`, but the API is otherwise modelled on `fetch()` and this failure is silent.

**Suggestion:** If `options.headers` is a `Headers` instance (or has `forEach`/`entries`), flatten it to a plain `{ name: value }` object before `cordova.exec`.

src/plugins/system/android/com/foxdebug/system/StreamHttp.java
```diff
@@ -0,0 +1,400 @@
...
+      cleanup();
+    }
+  }
+
+  private void waitForCredit(long bytes) {
+    synchronized (creditLock) {
+      while (!cancelled && pendingBytes + bytes > CREDIT_WINDOW_BYTES) {
+        try {
+          creditLock.wait();
+        } catch (InterruptedException e) {
```
bajrangCoder: **[suggestion]** On `InterruptedException`, `waitForCredit` sets the interrupt flag and then loops back into `creditLock.wait()`. `Object.wait()` throws immediately when the interrupt flag is already set, so an interrupt without `cancelled` becomes a tight spin until credit arrives or cancel is set. `cancel()` itself never interrupts the thread, so this is rare (runtime teardown), but it is still a busy-loop.

**Suggestion:** Treat interrupt as cancellation (`cancelled = true; return;`) or clear the interrupt flag before waiting again if the stream should continue.

src/plugins/system/android/com/foxdebug/system/StreamHttp.java
```diff
@@ -0,0 +1,400 @@
...
+      inputStream = stream;
+
+      if (stream != null) {
+        byte[] buffer = new byte[chunkSize];
+        int read;
+        while (!cancelled && (read = stream.read(buffer)) != -1) {
+          if (read <= 0) continue;
+          waitForCredit(read);
+          if (cancelled) break;
+          byte[] chunk = read == buffer.length
```
bajrangCoder: **[suggestion]** Every full-size read allocates a replacement `byte[chunkSize]`, and every partial read does `Arrays.copyOf`. `sendData` immediately encodes the bytes into a `String` / base64 and does not retain the array, so both the copy and the swap are unnecessary.

**Suggestion:** Reuse a single buffer and encode with the offset/length overloads: `new String(buffer, 0, read, StandardCharsets.ISO_8859_1)` and `Base64.encodeToString(buffer, 0, read, Base64.NO_WRAP)`.

tests/unit/systemHttpStream.test.js
```diff
@@ -0,0 +1,556 @@
...
+			if (done) break;
+			received.push({ time: Date.now() - start, chunk: value });
+		}
+
+		const text = received
+			.map((r) => Buffer.from(r.chunk).toString("utf8"))
+			.join("");
+		expect(text).toBe('data: {"token":"Hel"}\n\ndata: {"token":"lo"}\n\n');
+		expect(received.length).toBeGreaterThanOrEqual(2);
+		expect(received[0].time).toBeLessThan(280);
```
bajrangCoder: **[suggestion]** Test 2 asserts the first SSE chunk arrives in under 280ms and the last at ≥500ms, using wall-clock `Date.now()` against 300ms server delays. That is load-sensitive and can flake on a contended CI runner even when streaming is genuinely incremental.

**Suggestion:** Assert ordering/content of at least two chunks (already done) and relax or replace the absolute timing bounds, e.g. `received[0].time < received[received.length - 1].time` plus a looser lower bound, or inject a fake clock.

src/plugins/system/system.d.ts
```diff
@@ -296,6 +308,21 @@ interface System {
...
+   * only transport failures reject the promise. Cancelling the returned
+   * stream's reader (or aborting `options.signal`) cancels the underlying
+   * native HTTP request.
+   *
+   * @param url Request URL
+   * @param options Request options
+   * @returns A `Response` whose `body` is a `ReadableStream` of `Uint8Array` chunks
+   */
+  httpStream(url: string, options?: HttpStreamOptions): Promise<Response>;
+  /*
```
bajrangCoder: **[nit]** Inserting `httpStream` accidentally turned the following `setAppIcon` JSDoc from `/**` into `/*`, so it is no longer attached as documentation.

**Suggestion:** Restore `/**` on the `setAppIcon` comment.

src/plugins/system/android/com/foxdebug/system/System.java
```diff
@@ -656,6 +680,52 @@ private String getMimeTypeFromExtension(String fileName) {
     return mimeType != null ? mimeType : "application/octet-stream";
   }
 
+  /**
+   * Starts a streaming HTTP request. The response body is delivered to
+   * JavaScript incrementally as raw (or base64 encoded) chunks.
+   *
+   * <p>Args:
+   * <ol>
+   *   <li>requestId - unique id used for pause/resume/cancel</li>
```
bajrangCoder: **[nit]** The `httpStreamStart` javadoc says `requestId` is used for "pause/resume/cancel", but there is no pause/resume action — only `http-stream-ack` and `http-stream-cancel`.

**Suggestion:** Describe it as the id used for ACK/cancel.

_typos.toml
```diff
@@ -56,3 +56,4 @@ shft = "shft"
 multline = "multline"
 sav = "sav"
 ist = "ist"
+Hel = "Hel"
```
bajrangCoder: **[nit]** `Hel = "Hel"` is added to the global word list solely so the SSE fixture `"token":"Hel"` passes the typos linter.

**Suggestion:** Prefer a file-scoped ignore (or change the fixture token) instead of a repo-wide dictionary entry for a three-letter string.

bajrangCoder (approved):

PR #2874, opened 2026-09-10
feat(devcontainer): Add SSH feature in devcontainer.json

maaz7409:
Currently, we can't connect to codespace with configuration as per  `.devcontainer/*` as it lacks ssh feature in `devcontainer.json`.

Adding this enables ssh connection.

IMG_20260910_185831


RohitKushvaha01:
ssh here is unessesary

PR #2876, opened 2026-09-10
fix(android): reduce selection flicker

### LINKED ISSUE #2865: The screen flickers and goes black when selecting code.
### Check for existing issues

- [x] Completed

See the video. This issue occurs when the cursor crosses lines while selecting code; it happens almost always and has persisted for several versions.

https://github.com/user-attachments/assets/0dd330b4-0911-4c5f-9c90-b54b02a73ba0

### Environment

Device Information:
WebView Info:
		Package Name: com.google.android.webview
		Version: 154.0.8037.0

App Info:
		Name: Acode
		Package Name: com.foxdebug.acode
		Version: 1.13.3
		Version Code: 1009

Device Info:
		Android Version: 16
		Manufacturer: OPPO
		Model: PJW110
		Platform: Android
		Cordova Version: 15.1.0

Screen Info:
		Width: 360
		Height: 804
		Color Depth: 24

Additional Info:
		Language: zh-TW
		User Agent: Mozilla/5.0 (Linux; Android 16; PJW110 Build/BP2A.250605.015; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/154.0.8037.0 Mobile Safari/537.36

_No response_

</pre></details>

Degrace15:
Fixes #2865.

Reduces repeated interaction-guard updates during code selection by throttling setTouched() to one call per animation frame.


bajrangCoder:
Thanks for looking into this. I don’t think the proposed change has a clear connection to #2865.

`interactionGuard.markActive()` only updates an `activeUntil` timestamp using `performance.now()`; it does not modify the DOM, styles, or rendering state. This PR also leaves the rest of the `selectionchange` handler - `getSelection()`, `getRangeAt()`, and `closest()` running for every event, so it only throttles the cheapest operation.

The `requestAnimationFrame` latch also changes the guard’s behavior: if rendering is delayed, subsequent calls are ignored and the 200 ms guard may expire while selection is still active.

Could you please provide:

- A before-and-after recording on the affected device/WebView version.
- Evidence that calls to `markActive()` are causing the flicker.
- A regression test for the throttling behavior.

Without that validation, the black editor surface shown in #2865 appears more likely to be a WebView/compositor or selection-layer issue, so I don’t think this should be merged.

PR #2878, opened 2026-09-11
fix(terminal): stop proot "can't sanitize binding" fd warnings

RohitKushvaha01:
proot canonicalizes every -b host path with realpath(3). The /proc/self/fd/N magic links only resolve when the descriptor points to a real file, so piped stdio (what Acode's ProcessBuilder uses) made realpath(3) fail and proot drop the binding with `can't sanitize binding "/proc/self/fd/N"`.

The [ -e ] guard passed those links because stat(2) follows the magic link to the underlying pipe/socket inode, so it never filtered them out. Replace it with a can_bind helper that resolves the link target the way realpath(3) does, and narrow the LSP-side proot-warning filters to only this benign message so real proot warnings are no longer swallowed.


bajrangCoder (approved):

RohitKushvaha01:
@greptile

PR #2881, opened 2026-09-12
perf(editor): index language modes instead of sorting on every lookup

bajrangCoder:
## Summary
`getModeForPath` copied and sorted every registered language mode on each call. Directory listing hits this via icon resolution for every file; the sidebar tree hits it again while rendering. Icon plugins that wrap `helpers.getIconForFile` pay the same cost.

This rebuilds a lookup index only when `addMode` / `removeMode` run (exact filenames, filename regexes in specificity order, longest extension suffix) and caches results by basename.

File-tree icons are still resolved at render time. Caching the `sortDir` icon class would skip plugins that replace `helpers.getIconForFile` (Material Icons and similar), so that shortcut is intentionally not used here.

A dedicated file/folder icon plugin API can replace those monkey-patches later. This PR does not add that API.

## Compatibility
Public plugin APIs are unchanged:

- `acode.editorLanguages.*`
- `acode.aceModes.*`
- `ace.require("ace/ext/modelist")`
- `helpers.getIconForFile` (still overridable)

`getModes()` still returns registration order. Later plugin registrations still win on equal specificity. Register/unregister invalidates the index immediately.

Parity-tested against the old sort-and-scan for built-in modes, Dockerfile case sensitivity, compound extensions, nginx/BUILD filename matchers, and plugin install/remove.

## Measurements
4000 lookups (js/ts/json/py/md/css/unknown + Dockerfile/nginx/CMakeLists):

| Path | Time |
| --- | --- |
| Old copy+sort | ~808ms |
| Indexed | ~7.7ms |
| Warm cache | ~5.4ms |

About 105× faster in unit tests. Calling `getIconForFile` twice per tree row is still cheap after this; the old sort was the hitch.

## Follow-up
Add a first-class file/folder icon plugin API so themes like Material Icons do not need to replace `helpers.getIconForFile`.


bajrangCoder:
@greptile

PR #2887, opened 2026-09-12
feat: add official fileIcons API and Icon pack setting

bajrangCoder:
## Summary

Adds an official plugin API for file and folder icon packs, plus a user setting to choose the active pack.

Plugins register packs with `acode.require("fileIcons")`. The loader binds that API to the loading plugin and supplies ownership automatically. Acode owns matching, loading, fallback, and rendering. Users pick the pack in **Settings → App settings → Icon pack**. Builtin stays the default and remains available while a plugin pack is loading.

This replaces the earlier draft/internal icon helpers with a stable public surface: `register`, `icon`, and `onChange`. Pack selection, catalog queries, and DOM refresh stay internal.

## User-facing

- New **Icon pack** setting under App settings (interface).
- Builtin pack is always present.
- A previously selected plugin pack is restored automatically when that plugin registers again after startup.
- File and folder icons update in the explorer, recents, search results, and other file lists.

## Plugin API

Capture `acode.require("fileIcons")` synchronously in the plugin main script, and keep that API for later async work. Do not call `require("fileIcons")` for the first time after an `await`. The same bound API is also available as `options.fileIcons` in `setPluginInit`. Omit `pluginId`; if supplied, it must match the loading plugin.

```js
const fileIcons = acode.require("fileIcons");

acode.setPluginInit(plugin.id, (baseUrl) => {
  registration = fileIcons.register({
    id: plugin.id,
    name: "My Icons",
    icons: `${baseUrl}icons/`,
    fileNames: { "package.json": "nodejs" },
    fileExtensions: { js: "javascript", ts: "typescript" },
    folderNames: { src: "folder-src" },
    folderNamesExpanded: { src: "folder-src-open" },
    folder: "folder",
    folderExpanded: "folder-open",
  });
});

acode.setPluginUnmount(plugin.id, () => registration.dispose());
```

- `register(pack)` publishes or atomically replaces a pack. It does not select it.
- `icon(resource)` returns a CSS class string for a file or folder.
- `onChange(listener)` notifies when the active pack or assets change, for plugins that draw their own file lists.

Example plugin: [acode-icon-plugin-example](https://github.com/Acode-Foundation/acode-icon-plugin-example)


bajrangCoder:
@greptile

bajrangCoder:
@greptile

Elitex07:
`split()` and `join()` in `matchExtension()` allocate multiple new arrays and strings on the heap for every single file evaluated. In large directories, this will trigger unnecessary Garbage Collection (GC) pauses and slow down the render cycle. Although, not that much but maybe worth noting? 🤔

We can achieve the exact same behavior with zero array allocations by using `indexOf()` and string slicing; if that keeps the current implementation intacted ofc.

bajrangCoder:
> `split()` and `join()` in `matchExtension()` allocate multiple new arrays and strings on the heap for every single file evaluated. In large directories, this will trigger unnecessary Garbage Collection (GC) pauses and slow down the render cycle. Although, not that much but maybe worth noting? 🤔
> 
> We can achieve the exact same behavior with zero array allocations by using `indexOf()` and string slicing; if that keeps the current implementation intacted ofc.

Good point, I have fixed it

Elitex07 (approved):

Jeep654 (approved):

PR #2898, opened 2026-09-14
feat: plugin quicktool adapter

deadlyjack:
- allow plugin to adapt quicktool
- allow plugin to hook save, save event
- fix icon size in file explorer
- feat: icon button tooltip


deadlyjack:
@greptile review

src/handlers/quickTools.js
```diff
@@ -290,6 +381,38 @@ export default function actions(action, value) {
 
 	const { editor } = editorManager;
 	const { $input, $replaceInput } = quickTools;
+	const routed = quickToolsAdapters.has() && adapterAction(action, value);
+	if (routed) {
+		if (routed.type === "command" && routed.command === "openCommandPalette")
```
bajrangCoder: **[bug]** Host commands that should run on the active `EditorFile` are not special-cased the way `openCommandPalette` is. With an adapter registered, `actions("command", "saveFile")` becomes `{ type: "command", command: "saveFile" }` and is either dispatched into `adapter.execute` or dropped when `canHandle` is false (`return false` at line 391). It never reaches `acode.exec("save")` / `tab.save()`. A plugin that only follows `docs/file-save-events.md` (hook `tab.on("save")` and set `canSave`) therefore gets a working File menu Save, but the stock quicktools Save button no-ops or requires a second, easy-to-get-wrong `saveFile` implementation. Loose `canHandle: () => true` adapters also swallow Save without writing. Tests never click `data-value="saveFile"` on an adapted tab.

docs/file-save-events.md
```diff
@@ -0,0 +1,29 @@
+# Custom-tab Save and Save As
+
+Custom tabs use Acode's standard Save controls through the `EditorFile` save event. The plugin owns serialization, destination dialogs, writing and dirty tracking.
```
bajrangCoder: **[bug]** The new save API claims custom tabs use Acode's standard Save controls, and `canSaveFile` (used by the File menu) now accepts `file.canSave === true`. Keyboard Save does not. `saveFile` / `saveFileAs` remain `editorOnly: true` in `src/lib/keyBindings.js`, so the terminal shortcut path skips them. `src/handlers/keyboard.js` also returns early for `HTMLInputElement`, `HTMLTextAreaElement`, `HTMLSelectElement`, and `contenteditable` instead of forwarding Ctrl+S into the command that now calls `activeFile.save()`. Plugin editors built on textarea/contenteditable (the common case) therefore never hit `EditorFile.#save`. Non-input custom content only works accidentally: the document handler synthesizes a keydown on the hidden CodeMirror `contentDOM`, whose keymap then `acode.exec("save")`. If no editor view exists, even that path no-ops. Unit tests stub `file.save` and never drive Ctrl+S.

**Suggestion:** When `canSaveFile(activeFile)` and the tab is not a text editor, handle Ctrl+S / Ctrl+Shift+S in the document keyboard path (or drop `editorOnly` for those two bindings and stop ignoring save shortcuts inside the active custom tab). Do not depend on forwarding into CodeMirror.

src/handlers/quickToolsInit.js
```diff
@@ -102,7 +158,11 @@ export default function init() {
 	});
 
 	editorManager.on("editor-state-changed", updateHistoryButtons);
-	editorManager.on("switch-file", cancelQuickToolsModifierInput);
+	editorManager.on("switch-file", () => {
+		if (adapterWasActive || quickToolsAdapters.has()) cancelAdapterInput();
```
bajrangCoder: **[bug]** `switch-file` fires after `editorManager.activeFile` already points at the new tab. `cancelAdapterInput()` then calls `quickToolsAdapters.cancel()`, which aborts `current()` — the adapter you just switched to. The outgoing adapter is only aborted later in `sync()` via the stale `activeTab` pointer. Result: every time you focus an adapted tab, its `AbortController` is aborted, `pending`/`selection` are cleared, and `adapter.cancel()` runs (the documented `cancelPendingInput` hook). Incoming in-flight work is killed; plugins that cancel IME/selection on `cancel` will reset on tab show. The unit registry tests call `registry.sync()` directly and never go through this handler, so they miss it.

src/lib/quickToolsAdapter.js
```diff
@@ -0,0 +1,199 @@
...
+	const notify = (change) => listeners.forEach((listener) => listener(change));
+	const current = () => entries.get(getActiveTab());
+	function cancel(entry) {
+		if (!entry) return;
+		entry.controller.abort();
+		entry.controller = new AbortController();
+		entry.queue = Promise.resolve();
+		entry.pending = 0;
+		entry.selection = undefined;
+		entry.adapter.cancel?.();
```
bajrangCoder: `cancel(entry)` always invokes `entry.adapter.cancel?.()` with no re-entrancy guard. `register()`'s `subscribe` callback treats `busy`/`!enabled` as cancellation and calls `cancel(entry)` again. If a plugin's `cancel` reports busy (or `cancelAdapterInput` on switch causes a state notify), this recurses until the stack overflows. The docs tell plugins not to emit from `cancel`, but the host can still loop on a one-line mistake.

src/components/tooltip/longPress.js
```diff
@@ -0,0 +1,125 @@
...
+		}, 500);
+	}
+	function move(event) {
+		if (
+			press &&
+			Math.hypot(event.clientX - press.x, event.clientY - press.y) > 10
+		)
+			dismiss();
+	}
+	function up(event) {
```
bajrangCoder: **[suggestion]** `up()` clears the timer and arms click suppression but never calls `dismiss()`/`hide()`. After a long-press the tooltip stays at z-index 999999 until another pointerdown, scroll, keydown, or tab switch. The next tap is needed to dismiss it and is not suppressed (suppression is only for the original target), so that tap can activate a different control. `pointercancel` hides; ordinary release does not.

**Suggestion:** Call `hide()` (or `dismiss()` without clearing `suppressed`) in `up()` once `press.held` is true, so the tooltip disappears on finger lift and the 750ms click suppress still blocks accidental activation of the same icon.

src/lib/commands.js
```diff
@@ -206,10 +209,14 @@ export default {
 			strings["save all changes warning"],
 		);
 		if (!doSave) return;
-		editorManager.files.forEach((file) => {
-			file.save();
-			file.isUnsaved = false;
-		});
+		for (const file of [...editorManager.files]) {
+			if (!file.isUnsaved) continue;
+			if (!canSaveFile(file)) return false;
```
bajrangCoder: Save All / save-and-close now abort the whole run when they hit an unsaved tab with `canSaveFile` false (`return false` before remaining files). A dirty custom tab that never registered a save handler (or a plugin that set `isUnsaved` without `canSave`) blocks saving every text file behind it. Previously `file.save()` on non-editors resolved false and the loop continued (and Save All even forced `isUnsaved = false`). Stopping on a failed *savable* write is right; stopping on an unsavable sibling is a user-visible regression.

docs/file-save-events.md
```diff
@@ -0,0 +1,29 @@
+# Custom-tab Save and Save As
```
bajrangCoder: This is not the place for docs, docs should be in docs repo

deadlyjack:
@greptile review

deadlyjack:
@greptile review

src/lib/editorFile.js
```diff
@@ -1882,6 +1898,13 @@ export default class EditorFile {
...
 			if (this.uri) {
+				if (
+					!fsOperation.hasProvider(this.uri) &&
+					!isInitialPluginLoadComplete()
+				) {
+					await waitForInitialPluginLoad();
+				}
+				if (!this.#tab) return;
 				file ||= fsOperation(this.uri);
 				const fileExists = await file.exists();
```
bajrangCoder: Slow provider startup can delete recovery data

`waitForInitialPluginLoad()` resolves when the 15-second plugin-load race finishes, even though a timed-out plugin continues loading in the background. If its filesystem provider registers after that timeout, `fsOperation(this.uri)` is still undefined here and `.exists()` throws. The catch path removes the restored tab, and destruction can delete its cached unsaved content. Keep the tab pending until the provider registers, or fail without removing the tab/cache.

bajrangCoder (changes requested):

deadlyjack:
@greptile review

deadlyjack:
@greptile review

dyatwfyf-dot (approved):

bajrangCoder (approved):

PR #2913, opened 2026-09-19
Added: A Sample Logic for a mask.

dev12124:
I´m added a Mask using 0x5A for have a mask in the Critic Variables. e.g: I´m create a Variable using let named "secret_key" with attribute "secret_password". This Variable is in the RAM, and a Malware-Plugin installed have access this Variable and modify. And, e.g: The Variable loggedInUser have a Critic Properties ("email", "github", "website") and this Malware-Plugin access and view this Variable.


dev12124:
Hey @RohitKushvaha01, the masking issue pointed out by the Greptile bot has already been fixed in my latest commit (hash 2263642). 

I implemented a getter helper function called `getDecryptedUser` inside `auth.js` to dynamically unmask the properties right at the return statement of `getLoggedInUser`. This ensures the UI elements get their clean strings without crashing, while the global state variables remain obfuscated in the RAM heap via XOR. Could you please take a look at the updated files and consider reopening the PR? Thanks!

RohitKushvaha01:
> Hey @RohitKushvaha01, the masking issue pointed out by the Greptile bot has already been fixed in my latest commit (hash 2263642). 
> 
> I implemented a getter helper function called `getDecryptedUser` inside `auth.js` to dynamically unmask the properties right at the return statement of `getLoggedInUser`. This ensures the UI elements get their clean strings without crashing, while the global state variables remain obfuscated in the RAM heap via XOR. Could you please take a look at the updated files and consider reopening the PR? Thanks!
> 

XOR is not a encryption, it makes no difference in security also the things you are trying to make **safe** are not that sensitive, only sensitive thing worth securing is the access token which is already properly secured

PR #2915, opened 2026-09-19
fix(save): make format-on-save await the formatter and target the saved file

### LINKED ISSUE #2900: Format on Save can format the wrong tab and write before formatting completes
### Describe the bug

Format on Save uses the active editor instead of the file being saved. If the user switches from file A to file B while an asynchronous save step is pending, formatting can target B even though the save still writes A.

The save also starts formatting without awaiting completion. An asynchronous formatter can therefore finish after the source write, leaving the saved content unformatted. Formatter failure does not stop that write.

### Steps to reproduce

1. Enable Format on Save and configure a formatter for two editable files, A and B.
2. Edit A and use Save As with a provider whose destination `exists()` check is delayed. For deterministic reproduction, hold the final existence check after the destination/filename dialogs close and before formatting begins.
3. Switch to B before releasing that check.
4. Let the save continue: the formatter resolves the active file B, while the save retains A as its destination.
5. Separately, use an asynchronous formatter whose returned promise is delayed and observe whether the source write starts before formatting completes.

### Expected behavior

- Format and save the intended file, regardless of tab switches; leave other files untouched.
- Wait for formatting to complete before capturing and writing the document.
- Preserve dirty/cache state and cursor state on the correct file.
- Abort the save on formatting failure without reporting success.

### Cause and compatibility constraint

[`saveFile.js`](https://github.com/Acode-Foundation/Acode/blob/52d7d5d4504d5479bea7bbc63dbf0e083b1cd605/src/lib/saveFile.js#L125-L138) changes `editorManager.activeFile.markChanged` and calls `acode.exec("format", false)` without awaiting it. [`acode.format()`](https://github.com/Acode-Foundation/Acode/blob/52d7d5d4504d5479bea7bbc63dbf0e083b1cd605/src/lib/acode.js#L825-L865) selects the active file and invokes the formatter callback without a target argument.

Existing formatter plugins may read the active editor themselves, including after asynchronous work. Passing a file through command dispatch alone would not fix those callbacks. A later fix needs to account for compatibility, concurrent formatting requests, closure, and cleanup after failure; no implementation approach is selected here.

### Environment and scope

Identified by source review while working on #2898, at revision `52d7d5d4504d5479bea7bbc63dbf0e083b1cd605`. The steps above describe the controlled race; no specific released Android/WebView version is claimed as reproduced.

Track this separately from the cache/save fixes and defer implementation. Validate tab switching before and during formatting, the formatted source snapshot, and failure/closure behavior when addressing it.

### Check for existing issues

- [x] Searched existing open and closed issues; no matching report found.

xaniexane:
## Problem

Fixes #2900.

Format on Save had three races/corruption vectors in `src/lib/saveFile.js`:

1. **Wrong file formatted** — it toggled `editorManager.activeFile.markChanged` and ran the formatter against the active tab. If the user switched from file A to file B while an async save step (e.g. a slow provider `exists()` check during Save As) was pending, the formatter rewrote **B** while the save still wrote **A**.
2. **Write before format completed** — `acode.exec("format", false)` was fire-and-forget, so an async formatter could finish after the source write, leaving the saved content unformatted.
3. **Failure didn't stop the write** — a failed format still wrote the file and reported success.

## Fix

- `saveFile.js`: only run format-on-save when the file being saved is the active file (formatters operate on the active editor document by design — see below), toggle `markChanged` on the file being saved rather than whatever tab happens to be active, `await` the format so the write captures the formatted document, and abort the save when the formatter fails.
- `commands.js`: the `format` command now returns the format result so callers can react to it.
- `acode.js`: `acode.format()` returns `null` (instead of `false`) when no formatter is configured for the mode. Both are falsy for existing callers, but it lets save distinguish "nothing to do" (save proceeds) from "formatter failed" (save aborts) — so enabling format-on-save globally no longer breaks saving file types without a formatter.

## Known limitation (documented, not hidden)

The built-in formatters (Prettier, LSP) are view-bound: they read and rewrite the shared editor document, so they can only format the active tab. Truly formatting a background tab needs document-based formatting — a bigger architectural change. This PR makes the current architecture safe (no cross-file corruption, no phantom writes) rather than pretending to fix what it can't.

## Testing

- `biome check` clean on all touched files; `node --check` syntax-valid.
- Full `vitest` suite couldn't run here (no `node_modules` in this environment); the change is confined to the save path and the two existing `acode.format` call sites, both audited — no other callers exist.
- Verified by code inspection against the reproduction steps in #2900: with the guard, step 4 (formatter resolving file B) can no longer happen, and the un-awaited write in step 5 now awaits.

Happy to add unit coverage if there's a preferred mocking pattern for `editorManager`-dependent modules.


RohitKushvaha01:
@xaniexane, if you continue to submit untested slop PRs, you will be banned.

PR #2699, opened 2026-08-08
feat: replace alpine with ubuntu

RohitKushvaha01:
Closes #2200
Closes #2688

PR #2916, opened 2026-09-19
feat(build): add no-terminal flag to omit embedded terminal assets

### LINKED ISSUE #2897: CLI build option to omit embedded terminal assets
#### Context & Problem Statement
With the ongoing work to replace the Alpine terminal environment with Ubuntu (#2699), the embedded terminal rootfs installation payload increases from ~15 MB to roughly ~90 MB. This expands the overall compiled APK artifact size to approximately **113 MB**.

For developers working under a **capped/metered internet connection**, downloading ~113 MB APK artifacts for every incremental test run creates significant bandwidth overhead and friction during testing.

#### Developer Use Case & Workflow
* **Fork Testing Workflow**: Before opening a Pull Request or pushing updates to an active PR branch, test APKs are compiled via custom GitHub Actions workflows on personal forks to verify feature implementations and bug fixes.
* **Current CLI Limitations**: The standard build task `npm run build paid dev apk` forcefully bundles the complete embedded rootfs payload into the final APK.
* **F-Droid Build Workaround Constraints**: Although the `npm run build paid dev apk fdroid` command excludes certain assets, F-Droid builds target legacy Android SDK levels (API 28), making them unsuitable for accurately testing modern Android API behaviors and release builds.

#### Proposed Solution
Could we introduce an optional CLI flag or environment parameter (e.g., `--no-terminal-assets` or via `utils/storage_manager.mjs`) during the build step? 

This toggle would skip copying the heavy terminal rootfs package into `platforms/android/app/build/outputs/apk/debug/`, producing lightweight modern-target APKs specifically optimized for rapid iteration and bandwidth-constrained testing environments.

#### Key Benefits
* **Bandwidth Savings**: Drastically reduces artifact download sizes from ~113 MB down to a fraction of the size for developers on capped data plans.
* **Faster CI/CD Iteration**: Speeds up workflow execution, artifact packaging, and step summary generation during fork testing.
* **Targeted Verification**: Allows thorough testing of non-terminal app features against modern API targets without carrying the extra ~90 MB terminal payload.

<sub><sub>AI: Gemini 3.6 Flash</sub></sub>

xaniexane:
## Problem

Fixes #2897.

With the Alpine → Ubuntu terminal migration (#2699), the embedded rootfs payload grows from ~15MB to ~90MB, pushing test APKs to ~113MB. For developers on capped/metered connections, every fork-test build is a heavy download, and the F-Droid build workaround targets legacy API 28 — unsuitable for testing modern behavior.

## Change

Adds a `no-terminal` argument to the build script:

```
npm run build paid dev apk no-terminal
```

When set, the build skips the `com.foxdebug.acode.rk.exec.proot` plugin (which bundles the per-arch `alpine.rootfs` assets), so the rootfs is never copied into the APK. This follows the exact pattern the `fdroid` flag already uses for plugin add/remove, and a subsequent normal build re-adds the plugin automatically — no manual cleanup.

Deliberately untouched:
- Runtime `$FDROID` handling — a `no-terminal` build behaves like a normal build at runtime; the terminal simply fails fast on open (existing extract-asset error path) since its payload isn't bundled.
- The `fdroid` flag semantics — combining both flags still works (fdroid wins on runtime behavior).

## Testing

- `bash -n` syntax check passes.
- Exercised the argument/plugin logic in isolation with a stubbed `cordova` across all four states:
  - `no-terminal` + plugin present → plugin removed, notice printed
  - normal + plugin present → no plugin churn
  - `no-terminal` + plugin absent → no-op
  - normal + plugin absent → plugin re-added (self-healing)
- Full APK build not run here (needs Android SDK/Cordova toolchain); the change only gates plugin add/remove, which is the same mechanism `fdroid` has used in this script.


xaniexane:
Thanks for the review! One correction: the "Tab switch corrupts content" finding doesn't apply to this branch as it exists now. It was written against commit `9f01cc3`, which still contained the format-on-save work — that has since been moved to its own branch (PR #2915). This branch is now exactly `7d828e9`, which only touches `utils/scripts/build.sh` to add the `no-terminal` flag. No editor, formatter, or save-path code exists on this branch, so the finding can't apply here. No code change needed on this PR.

